PR #1, opened 2019-03-27
Update README.md

jboner:


jboner:
Added something more for a README. Probably needs to be evolved/changed though.

PR #31, opened 2019-04-29
Fixed entity key serialization

jroper:


src/backend/core/src/main/proto/protocol.proto
```diff
@@ -10,6 +10,8 @@ package com.lightbend.statefulserverless.grpc;
 import "google/protobuf/any.proto";
 import "google/protobuf/empty.proto";
 import "google/protobuf/descriptor.proto";
+// This is needed to ensure that the FieldOptions are extended when this file is loaded.
+import "entitykey.proto";
```
jroper: This is the fix. I had this in my local version of the copied file, which is why it worked for me.

Languages like JavaScript use the extensions to know what fields to include in the output, otherwise they are ignored. So, anything loading `protocol.proto` needs to ensure that `entitykey.proto` is also loaded to ensure it knows about the entity key.

src/backend/core/src/main/scala/com/lightbend/statefulserverless/Serve.scala
```diff
@@ -41,11 +42,33 @@ object Serve {
...
+    fieldOptions.withUnknownFields(scalapb.UnknownFieldSet(fields))
+  }
+
   private final class CommandSerializer(commandName: String, desc: Descriptor) extends ProtobufSerializer[Command] {
     private[this] final val commandTypeUrl = AnyTypeUrlHostName + desc.getFullName
-    private[this] final val entityKeys = desc.getFields.asScala.filter(
-      _.getOptions.getUnknownFields.hasField(EntityKeyOptionNumber) || true // FIXME require option to be set
-    ).toArray.sortBy(_.getIndex)
+    private[this] final val entityKeys = desc.getFields.asScala
+      .filter(field => EntitykeyProto.entityKey.get(convertFieldOptions(field)))
```
jroper: Much nicer entity key parsing, but does require the ugly conversion above to work.

src/samples/js-shopping-cart/test/test.js
```diff
@@ -187,6 +188,11 @@ describe("shopping cart", () => {
       service.should.have.property("name", "ShoppingCart");
       const GetShoppingCart = spec.root.lookupType("com.example.shoppingcart.GetShoppingCart");
       GetShoppingCart.fields.should.have.property("userId");
+      // protobuf.js doesn't transfer extension options when converting the protobuf descriptor to a protobuf.js root
+      // object, so we have to extract the field out of the protobuf descriptor
+      const userIdField = spec.proto.messageType.find(mt => mt.name === "GetShoppingCart")
+        .field.find(f => f.name === "userId");
+      userIdField.options.should.have.property(".com.lightbend.statefulserverless.grpc.entityKey", true);
```
jroper: And now we have a test to ensure the key is in the spec returned over gRPC.

src/backend/core/src/main/proto/protocol.proto
viktorklang: :+1:

src/backend/core/src/main/scala/com/lightbend/statefulserverless/Serve.scala
viktorklang: Nice!

src/samples/js-shopping-cart/test/test.js
viktorklang: Awesome

PR #33, opened 2019-05-03
Adds an initial TCK project

viktorklang:
Alright, so now we have the capability to start backends and frontends and verify both interactions and semantics—of course lots of tests yet to be written, but now we can add and run the tests as normal.

running the `test` task in the build will also run the tck (the default configuration is to run the akka backend with the js example frontend but is of course configurable).

TODO is to launch the akka backend without running it through sbt, I guess we'd have to come up with some form of alternative packaging, but it might be something worthwhile as we add Knative Build integration?


viktorklang:
@jboner @jroper This is ready to merge, I've fixed some semantical problems with the js-example implementation.

PR #38, opened 2019-05-21
Wip filedescriptorset √

viktorklang:
Migrates the js support to use protoc-generated .desc files and serving those instead of FileDescriptorProto or FileDescriptorSet in the protocol. This because JS has poor support for handling .desc itself.
Migrates the backend to use FileDescriptorSet instead of FileDescriptorProto which allows for more featureful .proto definitions of the external interfaces of user functions—while also making the ServerReflection support in the backend more feature-rich since it can now faithfully reflect the external API.

User API is minimally affected since it is possible to attach the protoc invocation in the package.json (as shown in this PR). Only remaining "issue" is to have to install `protoc`—something we could look at automating.


jroper:
Here's an example of a library (that we don't want to use) that downloads the right protoc compiler for the platform in an npm post install script:

https://github.com/YePpHa/node-protoc

I suggest we do that, and we provide a [bin script](https://docs.npmjs.com/files/package.json#bin) so that we can replace:

```
"prestart": "protoc --include_imports --proto_path=proto --proto_path=../../backend/core/src/main/proto --descriptor_set_out=user-function.desc proto/shoppingcart.proto"
```

with something like:

```
"prestart": "stateful-serverless-compile proto/shoppingcart.proto"
```

viktorklang:
@jroper Sounds like a good idea—would you be willing to take a stab at it?

viktorklang:
@jboner @jroper Merging this

PR #56, opened 2019-08-01
Proof-of-concept: Configuration to get native-image to generate a bin…

viktorklang:
…ary for the cloudstate proxy


proxy/core/src/main/resources/dev-mode.conf
```diff
@@ -0,0 +1,5 @@
+# Configuration for using an in memory journal and snapshot store
+include classpath("in-memory.conf")
```
viktorklang: @jroper I've removed the application.conf now and so you'll have to supply a -Dconfig.file|resource|url when launching the binary, for instance: -Dconfig.resource=dev-mode.conf

PR #58, opened 2019-08-09
Docker native image support

jroper:
Builds the native image in a docker container, and packages it as a Docker container too.


build.sbt
```diff
@@ -114,7 +194,7 @@ def sharedNativeImageSettings = Seq(
       )
 
 lazy val `proxy-core` = (project in file("proxy/core"))
-  .enablePlugins(JavaAppPackaging, DockerPlugin, AkkaGrpcPlugin, JavaAgent, AssemblyPlugin, GraalVMNativeImagePlugin)
+  .enablePlugins(DockerPlugin, AkkaGrpcPlugin, JavaAgent, AssemblyPlugin, GraalVMPlugin)
```
viktorklang: We should be able to remove the AssemblyPlugin too now, right?

proxy/core/src/main/scala/io/cloudstate/proxy/CloudStateProxyMain.scala
```diff
@@ -100,6 +101,20 @@ object CloudStateProxyMain {
...
+      // hooks don't get executed (so no graceful leaving of the cluster). Worse, if the process
+      // is the entrypoint for a Docker container (ie, it has pid 1) then it won't respond to TERM
+      // at all, because Linux does not implement the default TERM handling if pid is 1, the result
+      // being that the process will be killed after the configured termination timeout. So, we we
+      // need to register a TERM signal handler.
+      Signal.handle(new Signal("TERM"), _ => System.exit(0))
+
+      // And may as well register INT (Ctrl+C) while we're at it
+      Signal.handle(new Signal("INT"), _ => System.exit(0))
+    }
```
viktorklang: :+1:

viktorklang:
Great work, @jroper! :+1:

build.sbt
jroper: The tck still uses it by default, ie in CI, I think.

viktorklang: That's true. I guess if we have a Docker-client in CI we could launch it using that?

jroper: Yes, though building the native image takes a *lot* of resources, on my laptop it takes approx 5 minutes with close to 100% of all cores/hyperthreads being used concurrently (4 cores x 2 hyperthreads). Also, it needs 2GB of RAM (it oomed when I limited it to 2GB). I wouldn't be surprised if travis can't do it. Perhaps we could allow still building the non native image docker images and use that.

PR #65, opened 2019-08-17
Replace onlyIfChanged with Tracked.inputChanged

eed3si9n:
Ref https://github.com/sbt/sbt/issues/4940


viktorklang:
Thanks @eed3si9n! :)

PR #75, opened 2019-09-03
Added JavaScript api docs to the build

jroper:
Also hopefully fixed Java API docs generation


java-support/src/main/java/io/cloudstate/javasupport/ClientActionContext.java
```diff
@@ -2,7 +2,7 @@
 
 /**
  * Context that provides client actions, which include failing and forwarding.
- * <p/>
+ * <p>
```
jroper: javadoc 11 apparently doesn't like empty tags.

PR #85, opened 2019-09-06
[fix] this might be a "test" leftover from 08084e2

marcellanz:
@viktorklang I'm pretty sure about _reply consistency_ now, this regression ensured it strongly :-)


viktorklang:
Ouch! Thanks for your patience while trying to identify that commit-error!

PR #98, opened 2019-09-12
Java CRDT docs

jroper:


docs/src/main/paradox/user/lang/java/crdt.md
```diff
@@ -1,6 +1,141 @@
...
+
+In addition to the CRDT, the constructor may accept a @javadoc[`CrdtCreationContext`](io.cloudstate.javasupport.crdt.CrdtCreationContext).
+
+## Handling commands
+
+Command handlers can be declared by annotating a method with @javadoc[`@CommandHandler`](io.cloudstate.javasupport.crdt.CommandHandler). They take a context class of type @javadoc[`CommandContext`](io.cloudstate.javasupport.crdt.CommandContext).
+
+By default, the name of the command that the method handles will be the name of the method with the first letter capitalized. So, a method called `getCart` will handle gRPC service call command named `GetCart`. This can be overridden by setting the `name` parameter on the `@CommandHandler` annotation.
+
+The command handler also can take the gRPC service call input type as a parameter, to receive the command message. This is optional, sometimes it's not needed, for example, our `GetCart` service call doesn't need any information from the message, since it's just returning the current state as is. Meanwhile, the `AddItem` service call does need information from the message, since it needs to know the product id, description and quantity to add to the cart.
```
viktorklang: Perhaps we should document what happens if there are missing command implementations, or if there are multiple with the same "name" in the annodation?

docs/src/main/paradox/user/lang/java/crdt.md
```diff
@@ -1,6 +1,141 @@
...
+
+The return type of the command handler must be the output type for the gRPC service call, this will be sent as the reply.
```
viktorklang: What happens if it isn't?

docs/src/main/paradox/user/lang/java/crdt.md
```diff
@@ -1,6 +1,141 @@
...
+
+The following shows the implementation of the `GetCart` command handler. This command handler is a read-only command handler, it doesn't update the CRDT, it just returns some state:
+
+@@snip [ShoppingCartEntity.java](/docs/src/test/java/docs/user/crdt/ShoppingCartEntity.java) { #get-cart }
+
+## Updating a CRDT
+
+Due to CloudStates @ref[take in turns approach](../../features/crdts.md#approach-to-crdts-in-cloudstate), CRDTs may only be updated in command handlers and @ref[stream cancellation callbacks](#responding-to-stream-cancellation).
```
viktorklang: What happens otherwise?

docs/src/main/paradox/user/lang/java/crdt.md
```diff
@@ -1,6 +1,141 @@
...
+
+Here's a command handler for the `AddItem` command that adds the item to the shopping cart:
+
+@@snip [ShoppingCartEntity.java](/docs/src/test/java/docs/user/crdt/ShoppingCartEntity.java) { #add-item }
+
+## Deleting a CRDT
+
+A CRDT can be deleted by invoking @javadoc[`CommandContext.delete`](io.cloudstate.javasupport.crdt.CommandContext#delete--). Once a CRDT is deleted, the entity will be shut down, and all subsequent commands for the entity will be rejected.
+
+Caution should be taken when deleting CRDTs - the Reference Implementation of the proxy needs to maintain tombstones for each CRDT deleted, so over time, if many CRDTs are created and deleted, this will result in not just running out of memory, but increased network usage as the tombstones still need to be gossipped through the cluster for replication.
```
viktorklang: Perhaps we should state what "good" use-cases there are for deletion?

viktorklang: Users may want to have TTL for CRDTs, i.e. shopping carts might be deemed invalid after a certain amount of time has passed, how would they achieve that?

docs/src/main/paradox/user/lang/java/crdt.md
jroper: You get a failure at runtime when the user function is started.

docs/src/main/paradox/user/lang/java/crdt.md
jroper: If two have the same name, the user function will fail to start, with an error saying `Multiple methods found for handling command of name foo: methodOneName, methodTwoName`. If you define a method for a command that doesn't exist, you get this message: `Command handler method someMethodName for command SomeCommandName found, but the service has no command by that name.` And if a method is missing, you get this error message: `No command handler found for command SomeCommandName on CRDT entity: some.ClassName`. I don't think this needs to be documented, the error messages are completely clear, there's nothing the user needs to do that they couldn't work out themselves from the error messages, the documentation already describes the correct way of configuring it, there's no need for them to anticipate the errors ahead of time (eg they can't write any code to recover from it, the user function just refuses to start), and a single test that starts the entity will identify all the errors (though we do need to write a testkit so they can do this without a full cloudstate install).

docs/src/main/paradox/user/lang/java/crdt.md
jroper: Already raised an issue requesting the TTL feature: https://github.com/akka/akka/issues/27683

Currently, I don't think there are any good uses cases for deletion, it's just a limitation of Akka ddata that deletion is almost never a good idea. I did think about whether we should switch away from using top level CRDTs in the proxy, and instead shard them across ORMaps, that would solve the issue and make deletion possible. There's also a number of Akka ddata features that I've thought about, primarily, allowing entities to be passivated, so they don't get gossipped frequently, or even better, passivated to durable storage so they can be completely removed from memory. This would make deletion practical.

jroper: Here's an issue that would make deleting CRDTs feasible: https://github.com/akka/akka/issues/27714

viktorklang:
@jroper 
Failing check: Build code format check (address by running `sbt scalafmtSbt`)

PR #101, opened 2019-09-13
Fixed protobuf package import in load generator

bpg:
I guess this update simply got missed from a595d55f

Question to maintainers, should `load-generator` (well, any extra sources) be added as a _compile-time_ dependency to the project's root? Helps to catch errors when rarely used code simply falls of the radar...


viktorklang:
Thanks @bpg!

viktorklang:
@bpg raise the load-generator as a separate Issue, thanks!

PR #103, opened 2019-09-16
feature/go-support

### LINKED ISSUE #3: Golang support

marcellanz:
This PR implements Go support as a CloudState API Client Library and should fix #3 to a good amount.

Included is
- The implementation of the Discovery and EventSourced Services of CloudState.
- The implementation of the _TCK Shopping Cart_ example, used to be verified by the TCK.
- build support within travis-ci so that the TCK can validate the shopping-cart example (passes by 100%).
- a built.sbt task to build the Go binary with a Go Docker Container


viktorklang:
Thanks @marcellanz for opening this exciting PR!

I'm definitely no Go-expert so it's going to take some time to digest :)
Is it possible to avoid commiting the generated .pb.go-files?
Is it possible to avoid committing the .proto files for the protocol again, and instead refer to the files in protocols/?

Looking forward to having a look at this! :)

marcellanz:
@viktorklang 
:-) ahh... I opened the PR as a "Draft" to prepare it and its visible now :-o
But I'll fix some things your asked already (see below)

> Thanks @marcellanz for opening this exciting PR!
> I'm definitely no Go-expert so it's going to take some time to digest :)

"Go is simple and you should be able to learn it in a very short amount of time"(tm)
I'm curious about how well it is to read/understand for you, especially as you know the domain very well. As Go is a very simple (boring) language, it should be easy to follow. I hope its understandable. I tried my best to be explicit and idiomatic as possible as Go is usually explicit and has no magic going on. 

The most magic stuff happens with reflection; by which we need a bit of because of the dynamic nature of handlers. There is also one or two of sub-optimal things about the gRPC implementation; a ServiceDescriptor is not exposed explicitly by it and so it makes things like the registration of the entity and its service cumbersome.I iterated quite some times over that and hope I found something as expressive as possible without too much implicit magic. Actually the current form is explicit as a user can pass the proto filename; but also one of the proto message types used by the service.

> Is it possible to avoid commiting the generated .pb.go-files?
> Is it possible to avoid committing the .proto files for the protocol again, and instead refer to the files in protocols/?

I have both on an issue list on my branch here of what next steps for go-support might be here
https://github.com/marcellanz/cloudstate/issues/2

I'll clean up the proto file handling and its generation to be ready for this PR.

> Looking forward to having a look at this! :)

Thanks, I'm looking forward to your feedback.

viktorklang:
@marcellanz I took a stab at some of the build integration and proto generation things here: https://github.com/cloudstateio/cloudstate/commit/5b9df7c20b1da194d209f7a44b2bc29bdc601cd0

My Go is definitely not fluent enough :D I'll start looking at the Go-specific code shortly :)

I assume(?) that we want to split the go-support up into a library and the shopping-cart example, and put the go-shopping-cart in samples/… ?

I'm still not sure about the best way of running the compilation, if we should assume Go is installed or not, or run it in Docker.

marcellanz:
> @marcellanz I took a stab at some of the build integration and proto generation things here: [5b9df7c](https://github.com/cloudstateio/cloudstate/commit/5b9df7c20b1da194d209f7a44b2bc29bdc601cd0)

nice!

> My Go is definitely not fluent enough :D I'll start looking at the Go-specific code shortly :)
> 
> I assume(?) that we want to split the go-support up into a library and the shopping-cart example, and put the go-shopping-cart in samples/… ?

Yes, I'd support that and is what I had in my mind also to go forward. The *.proto files would be the only artefacts to share between libraries I think. "go imports" also expect that a go package lives on the edge of a VCS repository

> I'm still not sure about the best way of running the compilation, if we should assume Go is installed or not, or run it in Docker.

That was my motivation to introduce the _Docker supported go compilation_; so that the project as a whole can run the TCK and a CloudState developer doesn't need all compilers (Go, "node", Swift,...) installed correctly, which might be tricky, for the project. IMO the docker approach is kind of elegant to liberate the developers environment from that burden.

Having a library of a language support implementation would mean a separate repository, yes? It would be the most effective way to go for a Go library as Go's module and package management expects to use VCS endpoints for packages and imports directly.
I don't know too much about how a TCK should be structured, but it could even be outside of the main cloudstate repository and be together with something that builds "cloudstate + (any user function language support)" is needed. The CloudState "reference implementation" would be always a JVM based one; so keep these implementations together would be natural for a JVM based reference implementation of CloudState.

I have also not thought too much about the topic, but it came to me as I would have to provide a canonical "godoc" story which assumes to have a separate repository for the go code.

viktorklang:
@marcellanz 

>Yes, I'd support that and is what I had in my mind also to go forward. The *.proto files would be the only artefacts to share between libraries I think. "go imports" also expect that a go package lives on the edge of a VCS repository

I'd be totally open to have a cloudstateio/go-support repository, perhaps we should consider this for all language supports, at least when the CloudState Specification is more formalized, as there is less maintenance burden due to making sweeping changes across the board.

>That was my motivation to introduce the Docker supported go compilation; so that the project as a whole can run the TCK and a CloudState developer doesn't need all compilers (Go, "node", Swift,...) installed correctly, which might be tricky, for the project. IMO the docker approach is kind of elegant to liberate the developers environment from that burden.

Indeed. I'm not sure how we'll deal with Linux/Mac/Win support for that though. We'll need to figure out the testing matrix at some point.

>Having a library of a language support implementation would mean a separate repository, yes? It would be the most effective way to go for a Go library as Go's module and package management expects to use VCS endpoints for packages and imports directly.

I'm not sure whether it'd be strictly required or not. There are pros and cons to both models—monolithic builds are easier to overview etc, but having a repo for each language support makes for a cleaner interface between each.

>I don't know too much about how a TCK should be structured, but it could even be outside of the main cloudstate repository and be together with something that builds "cloudstate + (any user function language support)" is needed. The CloudState "reference implementation" would be always a JVM based one; so keep these implementations together would be natural for a JVM based reference implementation of CloudState.

Ideally we should be able to create a native executable using GraalVM native-image to build the TCK and put it in a Docker container for easy reuse, then it'd only be a matter of parameterizing a docker run to verify an implementation.

>I have also not thought too much about the topic, but it came to me as I would have to provide a canonical "godoc" story which assumes to have a separate repository for the go code.

Good point, I was not aware of this :)

@jroper or any other community member might want to weigh in? :)

marcellanz:
@viktorklang 
> > I have also not thought too much about the topic, but it came to me as I would have to provide a canonical "godoc" story which assumes to have a separate repository for the go code.
> 
> Good point, I was not aware of this :)

This, the godoc, should drive such a decision for sure and should be a priority. It can be done without it but it is more complex than I thought.

There is an old and still open issue about that on Go's side https://github.com/golang/go/issues/2381#issuecomment-529919077 where it is still surprisingly non-trivial to produce, just some "bunch of HTML godoc" for a package.

jroper:
Most of the comments I had have already been covered by @viktorklang.

How well does this work with `GOPATH`, does it require any hacks there or defining a custom one? Go's requirement for everything being inside a `GOPATH`, and in a very particular layout within there at that, has always struck me as being fundamentally incompatible with polyglot repositories, so I'm surprised that you've managed to get anything to compile at all. My assumption has always been that for this reason go support would have to have its own repo.

And it would probably make sense to give all the other the other languages their own repo, that said, given that everything is still under active development, that would introduce a lot of friction, so for now we should keep as much in this repo as possible, with the exception being go because as discussed.

If it's in a separate repo, would it make sense to download the proto files from a tag of this repo when the project is built? As I understand it, go package management tends to just clone dependencies from GitHub and other code repositories anyway, could that be leveraged to bring in the proto files?

marcellanz:
> Most of the comments I had have already been covered by @viktorklang.
> 
> How well does this work with `GOPATH`, does it require any hacks there or defining a custom one? Go's requirement for everything being inside a `GOPATH`, and in a very particular layout within there at that, has always struck me as being fundamentally incompatible with polyglot repositories, so I'm surprised that you've managed to get anything to compile at all. My assumption has always been that for this reason go support would have to have its own repo.

No hacks. Using Go Modules/go.mod helps in this case. Using it, GOPATH is nearly invisible or no issue at all as one can define the name of the module `module cloudstate.io/gosupport` like in
https://github.com/marcellanz/cloudstate/blob/feature/go-support/go-support/go.mod#L1

Before Go Modules this would have to be a "bit" more complicated, as `go get` would like to resolve any references to the _remote_ import path `cloudstate.io/gosupport` especially as we don't have this path public accessible at the moment. Without it, as you stated, it would not build so easily with a remote path.

Deciding which module path the go-support module would have is one item on my todo list
Either `cloudstate.io/gosupport` or `github.com/cloudstateio/gosupport` one of those [needs to have a VCS backed the source](https://golang.org/cmd/go/#hdr-Remote_import_paths). We could use local import paths, but I think this might not be a valid option for external users.

> And it would probably make sense to give all the other the other languages their own repo, that said, given that everything is still under active development, that would introduce a lot of friction, so for now we should keep as much in this repo as possible, with the exception being go because as discussed.

I agree, also regarding the friction. Having one of them outside could be a good test how to work with a separate repo. We even can let it be like it is right now as things build/work for now.

> If it's in a separate repo, would it make sense to download the proto files from a tag of this repo when the project is built? As I understand it, go package management tends to just clone dependencies from GitHub and other code repositories anyway, could that be leveraged to bring in the proto files?

That would be a viable way to go I think.

marcellanz:
@viktorklang 
> Is it possible to avoid commiting the generated .pb.go-files?
> Is it possible to avoid committing the .proto files for the protocol again, and instead refer to the files in protocols/?

These two are fixed and merged to the PR branch.

The command(s) to build the Go TCK part grew a bit, but I saw you have [revised the SBT task](https://github.com/marcellanz/cloudstate/blob/feature/go-support/.travis.yml#L61-L63). I think after this is merged too, the Go binary can be built also using the docker container.

viktorklang:
@marcellanz You can cherry-pick the changes in my branch to get the updates to the build etc.

marcellanz:
- [ ] cherry-pick the changes in my branch to get the updates to the build etc.

marcellanz:
> You can cherry-pick the changes in my branch to get the updates to the build etc.

@viktorklang I picket your change and on my branch as well as yours that [does not build properly](https://travis-ci.com/marcellanz/cloudstate/builds/128177587) on travis.

Screenshot 2019-09-18 at 15 39 53

are you able to spot why? I don't see from the two changes in build.sbt why that should fail.

viktorklang:
Try to run "sbt scalafmtSbt" and then commit the reformatted build file,
that should do the trick!

marcellanz:
> Try to run "sbt scalafmtSbt" and then commit the reformatted build file, that should do the trick!

oh did that it worked; I only had to read the error message but was blinded a bit by the expectation that it should not fail. thanks.

akramtexas:
@marcellanz - (1) I'm reviewing this PR and appreciate how you've left the history (aka `individual commits`) intact instead of `squash`ing them; this way, I can easily see how the chain of commits have evolved at a granular level (2) Question: Which version of `Go` are you using? Is it `1.12.4` by any chance?

marcellanz:
> @marcellanz - (1) I'm reviewing this PR and appreciate how you've left the history (aka `individual commits`) intact instead of `squash`ing them; this way, I can easily see how the chain of commits have evolved at a granular level (2) Question: Which version of `Go` are you using? Is it `1.12.4` by any chance?

@akramtexas 1) I usually let it this way, you can also find the quite hacky first PoC with which I started. 2) I decided to start with 1.13 right away. Thank you for reviewing it.

akramtexas:
- I second the option of having each specific language support library (e.g. a `cloudstateio/go-support` repository) live in its own repo as making the most sense. Among other things, and as @viktorklang succinctly pointed out: _"there is less maintenance burden due to making sweeping changes across the board."_ IMHO, managing repos via the strategy of "divide-and-conquer" can't be beat.
- Regarding the consideration of whether we assume Go is (natively) installed (versus run it in Docker), I lend my support to `Docker-supported Go compilation`. To add anecdotal credence to my statement, another open source Linux Foundation project that I've been immersed in for the past 14 months (it's called [EdgeX Foundry](https://www.edgexfoundry.org/)) relies on Docker-supported Go compilation, which is an option that continues to work efficiently and reliably. Given that this options works well for [Go projects](https://github.com/edgexfoundry) that I've worked on (with hundreds of thousands of LOC) gives me confidence that it should work equally well (or perhaps even better) for `CloudState`, which has a (relatively) modest amount of Go code to compile.

akramtexas:
Here's a snippet, FWIW, from [a sample](https://github.com/akramtexas/edgex-go/blob/master/cmd/sys-mgmt-agent/Dockerfile) `Dockerfile` (from the open source Linux Foundation project I referred to above) to illustrate what I have in mind:
`
RUN apk update && apk add make bash git
COPY go.mod .
# COPY go.sum .

RUN go mod download

COPY . .

# Build the SMA executable.
RUN make cmd/sys-mgmt-agent/sys-mgmt-agent

# Build the golang "executor" executable.
RUN make cmd/sys-mgmt-executor/sys-mgmt-executor

# Get the Docker-in-Docker image layered-in now.
FROM docker:latest

RUN apk add --no-cache bash
RUN rm -rf /var/cache/apk/*
...
...
# Copy over the SMA executable bits.
COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/sys-mgmt-agent /
COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/docker/configuration.toml
COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-agent/res/docker/configuration.toml /res/configuration.toml

# Copy over the golang "executor" executable.
COPY --from=builder /go/src/github.com/edgexfoundry/edgex-go/cmd/sys-mgmt-executor/sys-mgmt-executor /

RUN apk --no-cache add py-pip
RUN pip install docker-compose==1.23.2
RUN apk --no-cache add curl

ENTRYPOINT ["/sys-mgmt-agent", "--registry","--profile=docker","--confdir=/res"]
`

akramtexas:
> > @marcellanz - (1) I'm reviewing this PR and appreciate how you've left the history (aka `individual commits`) intact instead of `squash`ing them; this way, I can easily see how the chain of commits have evolved at a granular level (2) Question: Which version of `Go` are you using? Is it `1.12.4` by any chance?
> 
> @akramtexas 1) I usually let it this way, you can also find the quite hacky first PoC with which I started. 2) I decided to start with 1.13 right away. Thank you for reviewing it. Please consider that the PR is in Draft mode, I have a few tasks I listed above to do. But I appreciate any insights of yours.

@marcellanz This works for me.
- As they, say, _A team_, and I recall here from memory one of the wise pontifications by industry luminary Grady Booch, _that sticks consistently to a strategy will invariably outperform any other team that doesn't_ 🏁 Let's do it. 
- I finally found time (in the wee hours of the morning) to review your Go code and will now add comments inline.
- Overall, great work. I like your Go style and idioms a lot!

go-support/shoppingcart/cmd/shoppingcart/shoppingcart.go
```diff
@@ -0,0 +1,193 @@
...
+}
+
+// Handle lets us handle events by ourselves.
+//
+// returns handle set to true if we have handled the event
+// and any error that happened during the handling
+func (sc *ShoppingCart) HandleEvent(event interface{}) (handled bool, err error) {
+	switch e := event.(type) {
+	case *domain.ItemAdded:
+		return true, sc.ItemAdded(e)
```
akramtexas: Excellent use of the method receiver (`sc *ShoppingCart`) here 🆒

go-support/shoppingcart/cmd/shoppingcart/shoppingcart.go
```diff
@@ -0,0 +1,193 @@
...
+		})
+	}
+	return cart, nil
+}
+
+func (sc *ShoppingCart) Snapshot() (snapshot interface{}, err error) {
+	cart := domain.Cart{
+		Items: make([]*domain.LineItem, 0),
+	}
+	cart.Items = append(cart.Items, sc.cart...)
```
akramtexas: Does this need to be variadic?

marcellanz:
> * As they, say, _A team_, and I recall here from memory one of the wise pontifications by industry luminary Grady Booch, _that sticks consistently to a strategy will invariably outperform any other team that doesn't_ 🏁 Let's do it.
> * I finally found time (in the wee hours of the morning) to review your Go code and will now add comments inline.
> * Overall, great work. I like your Go style and idioms a lot!

@akramtexas thank you very much for your time you kindly took to review this PR. I'll have time tomorrow evening to go through your review comments and I'll look forward to any discussions that arise from.

akramtexas:
@marcellanz - The pleasure is all mine 👍 
- While I remain very much a programmer with a decidedly `Scala/Java` mindset, my immersion in the `Go` ecosystem (over the past 14 months-plus) has been an enlightening experience in terms of the idioms I have, and continue to, pick up. 
- I hope my review comments help out. 
- Yes, likewise, as a fellow volunteer-contributor, I'm super-excited about helping make `CloudState` an even more awesome project 🚀

go-support/cloudstate/eventsourced.go
```diff
@@ -0,0 +1,400 @@
...
+	}
+}
+
+func (esh *EventSourcedHandler) registerEntity(ese *EventSourcedEntity) error {
+	esh.entities[ese.ServiceName] = ese
+	return nil
+}
+
+// see EventSourcedServer.Handle
+func (esh *EventSourcedHandler) Handle(server protocol.EventSourced_HandleServer) error {
```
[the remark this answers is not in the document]
marcellanz: I fully support consistent naming. I this case these names are generated by the gRPC compiler plugin `protoc-gen-go` which uses a mix of CamelCase and snake_case identifiers for messages and fields and AFAIK I can't much about it for now.

There is also [an issue](https://github.com/golang/protobuf/issues/555) that addresses this behaviour on the protobuf support project of Go

go-support/shoppingcart/cmd/shoppingcart/shoppingcart.go
marcellanz: how it can be _"implicitly handled by the Go runtime"_? – I think thats not possible.

[func append(slice []Type, elems ...Type) []Type](https://golang.org/pkg/builtin/#append) is variadic, but as far as I know, the type of the second argument, `Type`, has to be the element type of the first argument of append by the [language spec](https://golang.org/ref/spec#Appending_and_copying_slices). As `sc.cart` is of type `[]*domain.LineItem` I have to suffix the second argument with the ellipsis `...` to be _spread_ into multiple arguments. This way, I don't have to loop over `sc.cart`.

go-support/shoppingcart/cmd/shoppingcart/shoppingcart.go
marcellanz: :-)

go-support/cloudstate/cloudstate.go
```diff
@@ -0,0 +1,246 @@
...
+	descriptorProto, err := unpackFile(proto.FileDescriptor(filename))
+	if err != nil {
+		return fmt.Errorf("failed to registerFileDescriptorProto for filename: %s: %w", filename, err)
+	}
+	r.fileDescriptorSet.File = append(r.fileDescriptorSet.File, descriptorProto)
+	return r.updateSpec()
+}
+
+func (r *EntityDiscoveryResponder) registerFileDescriptor(msg descriptor.Message) error {
+	fd, _ := descriptor.ForMessage(msg) // this can panic, so we do here
```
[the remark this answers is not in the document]
marcellanz: To my regret, I left quite a few Fatal failures left, just so that I got the full attention during start of development for these situations. They're all moved into proper errors with an update to this PR branch.

I also share your concerns regarding exit via `log.Fatalf` or `panic()` a program except for very valid reasons at its startup where it can't start at all. Also libraries should not panic, except perhaps internally.

In this case `descriptor.ForMessage` does panic by itself and also does not visibly states that by its function name, like using MustXXX. So I decided to let it panic until I had a better way to handle it. I fixed it and recover it know and move it into an error `registerFileDescriptor`will return.

go-support/cloudstate/cloudstate.go
```diff
@@ -100,23 +100,24 @@ func (cs *CloudState) Register(ese *EventSourcedEntity, config DescriptorConfig)
 }
 
 // Run runs the CloudState instance.
-func (cs *CloudState) Run() {
+func (cs *CloudState) Run() error {
 	host, ok := os.LookupEnv("HOST")
 	if !ok {
-		log.Fatalf("unable to get environment variable \"HOST\"")
+		return fmt.Errorf("unable to get environment variable \"HOST\"")
```
akramtexas: Much more meaningful response here, and elsewhere in your latest update 👍

go-support/cloudstate/eventsourced.go
```diff
@@ -309,24 +323,25 @@ func (esh *EventSourcedHandler) handleCommand(cmd *protocol.Command, server prot
...
 			// TODO: protobufs are expected here, but CloudState supports other formats
 			message, ok := evt.(proto.Message)
 			if !ok {
-				log.Fatalf("got a non-proto message as event")
+				return nil, fmt.Errorf("got a non-proto message as event")
 			}
 			marshal, err := proto.Marshal(message)
 			if err != nil {
-				log.Fatalf("unable to Marshal")
+				return nil, fmt.Errorf("%s, %w", err, ErrMarshal)
```
akramtexas: Excellent use of multi-argument logging. Succinctly done 👍

akramtexas:
@marcellanz I have added a couple more comments this Sunday morning (based on your latest change-set push). Regarding my set of prior comments, I thank you for addressing them, plus I appreciated your detailed responses, including this one, which, incidentally, makes perfect sense:
> func append(slice []Type, elems ...Type) []Type is variadic, but as far as I know, the type of the second argument, Type, has to be the element type of the first argument of append by the language spec. As sc.cart is of type []*domain.LineItem I have to suffix the second argument with the ellipsis ... to be spread into multiple arguments. This way, I don't have to loop over sc.cart.
- I got your responses via Github-generated _emails_, but, for some reason, those responses don't appear in this PR (neither under the `Conversations` tab nor the `Files changed` tab)...

marcellanz:
@arkamtexas github hides "outdated"  comments. try to find "show outdated" buttons to show them. outdated means, updated by a changing commit.

go-support/cloudstate/eventsourced.go
```diff
@@ -0,0 +1,416 @@
...
+		return fmt.Errorf("failed to unmarshal: %w", err)
+	}
+
+	inputs[2] = reflect.ValueOf(msg)
+	// call it
+	called := method.Func.Call(inputs)
+	// The gRPC implementation returns the rpc return method
+	// and an error as a second return value.
+	errReturned := called[1]
+	if errReturned.Interface() != nil && errReturned.Type().Name() == "error" { // FIXME: looks ugly
```
marcellanz: calling .Interface() would panic if it cant; better call CanInterface()

go-support/cloudstate/eventsourced.go
```diff
@@ -0,0 +1,400 @@
...
+			// TODO: protobufs are expected here, but CloudState supports other formats
+			message, ok := evt.(proto.Message)
+			if !ok {
+				log.Fatalf("got a non-proto message as event")
+			}
+			marshal, err := proto.Marshal(message)
+			if err != nil {
+				log.Fatalf("unable to Marshal")
+			}
+			events = append(events,
```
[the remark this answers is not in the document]
marcellanz: >Did you have any concerns regarding the order in which events are emitted
@akramtexas Actually I'm not. This should be completely in sequence of what is stated instruction wise. Since theres is, at the first call shortly discussed, no concurrent data path involved, the EventSourcedHandler has at all times full control over the path how events are flowing around. Do you see any races here?

go-support/shoppingcart/cmd/shoppingcart/shoppingcart.go
akramtexas: That makes sense (I was thinking out loud to eliminate a possible source of flaw, but you have satisfied my potential concern, all good here) 👍

marcellanz: >I was thinking out loud to eliminate a possible source of flaw

You're most welcome. Please do at any time.

go-support/cloudstate/eventsourced.go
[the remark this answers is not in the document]
akramtexas: @marcellanz Here, too, you have removed my concern. Looks good. For me, the final word on concurrency (in `Go`!) is the fine book named _“Concurrency in Go”_ by Katherine Cox-Buday, where she has a helpful tip toward the end of the book. FWIW, like so:
```
“In Go 1.1, a -race flag was added as a flag for most go commands:
  $ go test -race mypkg    # test the package
  $ go run -race mysrc.go  # compile and run the program
  $ go build -race mycmd   # build the command
  $ go install -race mypkg # install the package
If you’re a developer and all you need is a more reliable way to detect race conditions, this is really all you need to know. One caveat of using the race detector is that the algorithm will only find races that are contained in code that is exercised. For this reason, the Go team recommends running a build of your application built with the race flag under real-world load. This increases the probability of finding races by virtue of increasing the probability that more code is exercised.”
```

go-support/cloudstate/eventsourced.go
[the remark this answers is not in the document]
akramtexas: Ah, I see. No worries in that case, we're good for now 👍

go-support/cloudstate/eventsourced.go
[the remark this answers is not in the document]
marcellanz: @akramtexas yeah the `-race` flag is of great help for detecting races in code. As a precaution I have been using this flag with the [`go-support/build/run_go_test_in_docker.sh`](https://github.com/cloudstateio/cloudstate/pull/103/commits/c1afd6ac4ff0423c2f74c7c89db37d7a12f59356#diff-b92d99f75509105ea6d3ff4d474113bdR7) script

`go test -count 1 -race ./...`

We might even run the TCK with the flag enabled to get covered here.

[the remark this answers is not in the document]
akramtexas: @marcellanz Good to hear about your routine use of the `-race` flag 👍 And yes, I support running the TCK with the flag enabled to get covered in this area of `Go` code, provided that it does not create any (infrastructural) burden.

viktorklang:
@marcellanz Great progress here! Would you prefer to have this under its own cloudstateio/go-support repository? I'm thinking we might want to start splitting language supports into their own repos rather soon-ish, to make sure that independent progress can be made. Let me know what you think!

marcellanz:
> @marcellanz Great progress here! Would you prefer to have this under its own cloudstateio/go-support repository? I'm thinking we might want to start splitting language supports into their own repos rather soon-ish, to make sure that independent progress can be made. Let me know what you think!

I would prefer it yes. 

The go related cloudstate documentation would stay within cloudstateio/cloudstate I think and the godoc's then can be served through cloudstateio/go-support via [godoc.org](godoc.org), the way Go likes to get api doc delivered. 

Then I'll have to figure how to build the TCK part, but it might be just a  `go get github.com/cloudstateio/go-support/shoppingcart` within the TCK build.

For any splitted language-support repo and said in #104 and like these two points above, the splitting affects documentation, tck builds and the sharing of the protocol with its *.proto files.

As we moved the protobuf files out of the go-support branch to have them not duplicated, they'll have to be somewhere for an build of the language repo. I'll look into that and what the options are.
Prefereably I'd not like to `_curl_` then during a build, or whats your preference on here?

viktorklang:
@marcellanz Ok, cool. I'll check with @jroper what we need to set up. CLA validator, CI, access rights to the repo etc.

viktorklang:
@marcellanz I created this now, I'll have to configure it soon-ish: https://github.com/cloudstateio/go-support

marcellanz:
> @marcellanz I created this now, I'll have to configure it soon-ish: https://github.com/cloudstateio/go-support

@viktorklang Cool. I assume this PR then is better not merged here. I'll fork it and prepare its separation from the main repo.

marcellanz:
@akramtexas as shortly discussed above I will close this PR in the coming days as I'm moving the code into a separate go-support repository. I will reference this then closed PR within a new PR.

The task is captured in [this issue](https://github.com/cloudstateio/go-support/issues/1) on the new repository.

Thank you for your review input so far. I'll looking forward to the initial PR on the go-support repo which will have all input of yours included; I think I will also rebase whats been done so far to have a clean history for the first commit.

akramtexas:
@marcellanz I, too, support that this PR is better off not being merged here, and instead going with the option of forking and separating (the work done here, to date) from the main repo ✅ 

I am happy to contribute (and will continue to do so) with review input based on my know-how of `Go` and its ecosystem 👍

marcellanz:
as in https://github.com/cloudstateio/go-support/issues/1 announcent, I close this PR as it will be moved to its own repository.

PR #107, opened 2019-09-23
Docs for effects in Java

jroper:


docs/src/test/java/docs/user/effects/ShoppingCartEntity.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class ShoppingCartEntity {
+
+  // #lookup
+  private final ServiceCallRef<Hotitems.Item> itemAddedToCartRef;
+
+  public ShoppingCartEntity(Context ctx) {
+    itemAddedToCartRef =
+        ctx.serviceCallFactory()
+            .lookup("example.shoppingcart", "ItemAddedToCart", Hotitems.Item.class);
```
viktorklang: If this method requires having the proto classes for the endpoint-to-be-called, is there any other way of referring to an endpoint rather than look up by Strings?

docs/src/test/java/docs/user/effects/ShoppingCartEntity.java
```diff
@@ -0,0 +1,58 @@
...
+  // #lookup
+
+  class CommandHandlerWithForward {
+    // #forward
+    @CommandHandler
+    public void addItem(Shoppingcart.AddLineItem item, CommandContext ctx) {
+      // ... Validate and emit event
+
+      ctx.forward(
+          itemAddedToCartRef.createCall(
```
viktorklang: An idea w.r.t. naming. How about `prepareCall`?

docs/src/test/java/docs/user/effects/ShoppingCartEntity.java
jroper: No. Not unless you've compiled the grpc descriptor to something. We don't require the gRPC descriptor to be compiled to anything in Java, since we dynamically serve it, we just read the descriptor directly, and then users implement it using annotations which are reflected over. By default, nothing will be generated for the gRPC services, other than the descriptor, and so they can only be looked up by string. If we added support for generating code, then we could in fact remove the need for the `ServiceCallFactory` at all, and instead emitting an effect could look like this:

```
ctx.effect(HotItemsService.ref.itemAddedToCart(item));
```

docs/src/test/java/docs/user/effects/ShoppingCartEntity.java
jroper: Agree it's not the best naming, but what exactly is being prepared? We're creating something that has a reference to a call and the message to pass to it - prepare sounds like something more than that is being done, eg perhaps resources are being allocated ready to make the call, which is not the case.

PR #116, opened 2019-09-30
scala support

janory:
### Steps:
- copy the java-support :white_check_mark:
- make it work as it is :white_check_mark:
- extract the common parts (as discusses in the meeting separate PR)
- change annotation based support, to more idiomatic  (as discusses in the meeting separate PR)
- add CRDT support  (as discusses in the meeting separate PR)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -627,6 +627,70 @@ lazy val `java-support` = (project in file("java-support"))
     )
   )
 
+lazy val `scala-support` = (project in file("scala-support"))
+  .enablePlugins(AkkaGrpcPlugin, BuildInfoPlugin)
+  .settings(
+    name := "cloudstate-scala-support",
+    common,
+    crossPaths := false,
+    publishMavenStyle := true,
+    publishTo := sonatypePublishTo.value,
+    buildInfoKeys := Seq[BuildInfoKey](name, version),
+    buildInfoPackage := "io.cloudstate.scalasupport",
+    // Generate javadocs by just including non generated Java sources
+    sourceDirectories in (Compile, doc) := Seq((javaSource in Compile).value),
+    sources in (Compile, doc) := {
+      val javaSourceDir = (javaSource in Compile).value.getAbsolutePath
+      (sources in (Compile, doc)).value.filter(_.getAbsolutePath.startsWith(javaSourceDir))
+    },
+    // javadoc (I think java 9 onwards) refuses to compile javadocs if it can't compile the entire source path.
+    // but since we have java files depending on Scala files, we need to include ourselves on the classpath.
+    dependencyClasspath in (Compile, doc) := (fullClasspath in Compile).value,
+    javacOptions in (Compile, doc) ++= Seq(
+      "-overview",
+      ((javaSource in Compile).value / "overview.html").getAbsolutePath,
+      "-notimestamp",
+      "-doctitle",
+      "CloudState Scala Support"
+    ),
+    libraryDependencies ++= Seq(
+      // Remove these explicit gRPC/netty dependencies once akka-grpc 0.7.1 is released and we've upgraded to using that
+      "io.grpc" % "grpc-core" % GrpcJavaVersion,
+      "io.grpc" % "grpc-netty-shaded" % GrpcJavaVersion,
+      "com.typesafe.akka" %% "akka-stream" % AkkaVersion,
+      "com.typesafe.akka" %% "akka-slf4j" % AkkaVersion,
+      "com.typesafe.akka" %% "akka-discovery" % AkkaVersion,
+      "com.typesafe.akka" %% "akka-http" % AkkaHttpVersion,
+      "com.typesafe.akka" %% "akka-http-spray-json" % AkkaHttpVersion,
+      "com.typesafe.akka" %% "akka-http-core" % AkkaHttpVersion,
+      "com.typesafe.akka" %% "akka-http2-support" % AkkaHttpVersion,
+      "com.google.protobuf" % "protobuf-java" % ProtobufVersion % "protobuf",
+      "com.google.protobuf" % "protobuf-java-util" % ProtobufVersion,
+      "org.scalatest" %% "scalatest" % ScalaTestVersion % Test,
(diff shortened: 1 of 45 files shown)


viktorklang:
Hmmm, weird CI failure, could be linked to Coursier: https://github.com/sbt/sbt/issues/4296 Will have to dig deeper.

janory:
> Hmmm, weird CI failure, could be linked to Coursier: [sbt/sbt#4296](https://github.com/sbt/sbt/issues/4296) Will have to dig deeper.

Yeah, it seems unrelated.

build.sbt
```diff
@@ -627,6 +627,70 @@ lazy val `java-support` = (project in file("java-support"))
     )
   )
 
+lazy val `scala-support` = (project in file("scala-support"))
```
viktorklang: Try addding this project to `aggregate`:

```
.aggregate(`proxy-core`,
             `proxy-cassandra`,
             `proxy-postgres`,
             `java-support`,
             `scala-support`, <--- HERE
             `java-shopping-cart`,
             `akka-client`,
             operator,
             `tck`,
             docs)
  .settings(common)
```

janory:
@viktorklang 

I am currently having an issue with the assembly of the `scala-shopping-cart`:

```
[error] stack trace is suppressed; run last scala-shopping-cart / assembly for the full output
[error] (scala-shopping-cart / assembly) deduplicate: different file contents found in the following:
[error] com/google/api/AnnotationsProto.class
[error] com/google/api/AnnotationsProto.class
[error] deduplicate: different file contents found in the following:
[error] com/google/api/CustomHttpPattern.class
[error] com/google/api/CustomHttpPattern.class
[error] deduplicate: different file contents found in the following:
[error] com/google/api/Http.class
[error] com/google/api/Http.class
[error] deduplicate: different file contents found in the following:
[error] com/google/api/HttpProto.class
[error] com/google/api/HttpProto.class
[error] deduplicate: different file contents found in the following:
[error] com/google/api/HttpRule.class
[error] com/google/api/HttpRule.class
```

My gut feeling says, that the `scala-support` module already generates these files and because `scala-shopping-cart` is depending on that, it is generated two times.
But if I remove the `baseDir / "frontend"` from the `scala-shopping-cart` the assembly also fails, because the classes are not generated at all.

build.sbt
```diff
@@ -656,6 +723,29 @@ lazy val `java-shopping-cart` = (project in file("samples/java-shopping-cart"))
...
 
+lazy val `scala-shopping-cart` = (project in file("samples/scala-shopping-cart"))
+  .dependsOn(`scala-support`)
+  .enablePlugins(AkkaGrpcPlugin)
+  .settings(
+    name := "scala-shopping-cart",
+    PB.generate in Compile := (PB.generate in Compile).dependsOn(PB.generate in (`scala-support`, Compile)).value,
+    PB.protoSources in Compile ++= {
+      val baseDir = (baseDirectory in ThisBuild).value / "protocols"
+      Seq(baseDir / "frontend", baseDir / "example")
```
viktorklang: @janory Have you tried

```suggestion
      Seq(baseDir / "example")
```

janory: @viktorklang 
I've checked out the branch & run the TCK on a different computer and it worked well, so I will cleanup the project and try it again. It is definitely a local issue.

The TCK is almost complete, the only issue is this:
```
[info] - must verify that the HTTP API of ShoppingCart protocol works *** FAILED ***
"...ems":[{"productId":"[B14623482","name":"Basic","quantity":1},{"productId":"A14362347","name":"Deluxe","quantity":7]}]}" did not equal "...ems":[{"productId":"[A14362347","name":"Deluxe","quantity":7},{"productId":"B14623482","name":"Basic","quantity":1]}]}" 
```

It is the same content, but the order is different. 
I will check the implementation and fix the order, although maybe we can relax the TCK's check to ignore the order and focus only the content, because my guess is, that the order here is not important.

marcellanz: @janory I had the excactly same "issue" in the Go Support first :-), had the items in a list that is not sorted.

janory: @marcellanz :)
It's a one liner fix, I am just not sure if we really need to enforce the order here.

marcellanz: Had the same thought, yes.
There are other implicit requirements by the shopping cart tck example. Adding negative amounts of line-items as an example. As the shopping cart is used to verify the spec through the TCK, we perhaps might document these requirements within the shoppingcart.proto file.

viktorklang: The reason for checking the response as a String was to avoid having parsing influencing things needlessly. I'm very much open for a better way of verifying the responses :)

viktorklang:
@janory Ok, so the next steps is to create a Scala-idiomatic API, and then once that's done, to see what commonalities still exist between the java and scala support, and pull those out into a jvm-support library?

janory:
> @janory Ok, so the next steps is to create a Scala-idiomatic API, and then once that's done, to see what commonalities still exist between the java and scala support, and pull those out into a jvm-support library?

@viktorklang exactly! But, I think those should happen in a separate PR.

build.sbt
```diff
@@ -630,6 +633,70 @@ lazy val `java-support` = (project in file("java-support"))
...
+        "com.google.protobuf" % "protobuf-java-util" % ProtobufVersion,
+        "org.scalatest" %% "scalatest" % ScalaTestVersion % Test,
+        "com.typesafe.akka" %% "akka-testkit" % AkkaVersion % Test,
+        "com.typesafe.akka" %% "akka-stream-testkit" % AkkaVersion % Test,
+        "com.typesafe.akka" %% "akka-http-testkit" % AkkaHttpVersion % Test,
+        "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf",
+        "org.slf4j" % "slf4j-simple" % "1.7.26",
+        "com.fasterxml.jackson.core" % "jackson-databind" % "2.9.9.3"
+      ),
+    javacOptions in Compile ++= Seq("-encoding", "UTF-8"),
```
viktorklang: Move this to the other javacOptiuons above?

viktorklang:
@janory Cool, let the next step begin! :)

Issue #97, opened 2019-09-11
.NET Core support


viktorklang:
It would be really cool to support .NET Core for user functions!

nagytech:
I've ported some of Lagom to NETcore in the past (nagytech/lagom-dotnet) so I'm keen to help with a NETcore implementation of CloudState.  

But, I've run into a few minor issues so far.  First, I've had to to dig into the NETCore gRpc implementation.  The current `Google.Protobuf` library they use protects the reflective Descriptor classes.  When I get to a point where it will build and I'm not embarrassed about the code, I'll check it in for some review.

viktorklang:
@nagytech Very cool, this will be very interesting! Let us know if you have any questions, the gitter channel is typically a good place to ask questions.

nagytech:
I've added my implementation here: https://github.com/nagytech/cloudstate-csharp/

But, as mentioned earlier I still need to resolve the issues in the csharp libraries for Protobuf and gRPC.  One of them will need to wait for Protobuf 3.11.0: https://github.com/protocolbuffers/protobuf/issues/5441

Additionally, there is an outstanding issue in VSCode that makes developing gRPC quite difficult: 

In the meantime, I'm going to refine and document what I've created.

viktorklang:
@nagytech Cool, I'll have a look!

sleipnir:
I believe that this can be merged with an official repository on cloudstate github account. I did a review with Jonathan that resulted in this minor [issue](https://github.com/nagytech/cloudstate-csharp/issues/6), but I believe that everything else is fit.

sleipnir:
The issue is not related to TCK testing features so it should not impact anyone using this library.

sleipnir:
> The issue is not related to TCK testing features so it should not impact anyone using this library.


but I still sent a PR https://github.com/nagytech/cloudstate-csharp/pull/7

viktorklang:
@sleipnir @nagytech Created this: https://github.com/cloudstateio/dotnet-support prepare PR?

nagytech:
Thanks @viktorklang.  I'll send one through in the next day or so.

viktorklang:
Perfect, thanks @nagytech!

PR #127, opened 2019-10-09
Add namespace option to for csharp protoc plugin

nagytech:
In support of #97 protoc needs to know about the `Google.Protobuf` C# namespace.


viktorklang:
Thanks @nagytech!

Issue #49, opened 2019-07-17
Support Evented services


viktorklang:
Being able to emit events to be consumed by other services/endpoints as a part of command processing, and then being able to declare/autogenerate that an endpoint is able to receive certain kinds of events, either by having a transformation from an event to a command, and use the existing gRPC endpoints, or having to implement a specific endpoint signature to be able to receive events of that kind.

Implementing support for this will require a bit of R&D

viktorklang:
One possibility is using CloudEvents for this.

viktorklang:
@akramtexas Please do! We can raise this topic in the inaugural contributors' confcall next week

akramtexas:
@viktorklang Awesome, thanks! I can't wait to throw myself at this 🆒 stuff.
- While doing a decidedly preliminary skimming-through of our source, I see some "hooks" already in place, for example this one in `/cloudstate/java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/AnnotationBasedEventSourcedSupportSpec.scala`:
`      "allow emiting events" in {
        val handler = create(new {
          @CommandHandler
          def addItem(msg: String, ctx: CommandContext) = {
            ctx.emit(msg + " event")
            ctx.commandName() should ===("Add Item")
            Wrapped(msg)
          }
        }, method)
        val ctx = new MockCommandContext
        decodeWrapped(handler.handleCommand(command("blah"), ctx).get) should ===(Wrapped("blah"))
        ctx.emited should ===(Seq("blah event"))
      }
`
...and this one `/cloudstate/java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/AnnotationBasedEventSourcedSupportSpec.scala`:
`            if (jsonClass.getAnnotation(classOf[Jsonable]) == null) {
              throw SerializationException(
                s"Illegal CloudEvents json class, no @Jsonable annotation is present: $jsonType"
              )
            }`
- Getting to be time to dig in ⛏

viktorklang:
@akramtexas That's for the eventsourcing parts. Those events are not necessarily intended for anything but internal consumption. Think about "Evented Services" as being able to consume and produce events coming in and going out externally. :)

akramtexas:
@viktorklang Gotcha, that makes sense. Plus I thank you for keeping me honest :)

viktorklang:
@akramtexas No problem :) Just want to spare you some time & effort :)

akramtexas:
[...]

PR #141, opened 2019-11-06
Adding initial, wip, support for Eventing in CloudState

viktorklang:
Addresses https://github.com/cloudstateio/cloudstate/issues/49

This is a Draft PR since it needs feedback,
I am currently working on solving so that tests can be added and run against the Google Pubsub emulator locally, especially in CI.

Also missing is:
Reference documentation


build.sbt
```diff
@@ -522,7 +531,8 @@ lazy val `proxy-tests` = (project in file("proxy/proxy-tests"))
   .settings(
     common,
     name := "cloudstate-proxy-tests",
-    fork in Test := true,
+    fork in Test := System.getProperty("RUN_STRESS_TESTS", "false") == "true",
```
viktorklang: @jroper I thought this sensible, since running stress tests on every commit/PR seems wasteful right now

viktorklang:
I've fixed a couple of defects and made it possible to run Google Pubsub Emulator on CI, so now we can add tests!

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/EventingManager.scala
```diff
@@ -0,0 +1,124 @@
...
+
+  final def serve(eventing: Eventing,
+                  commandHandler: CommandHandler,
+                  entityDiscoveryClient: EntityDiscoveryClient,
+                  log: Logger): Source[AnyRef, Future[Cancellable]] =
+    createSource(eventing.in, commandHandler)
+      .map(commandHandler.serializer.parse)
+      .via(commandHandler.flowUsing(entityDiscoveryClient, log))
+      .via(createDestination(eventing.out, commandHandler))
+      .dropWhile(_ => true) // TODO decide what to do with these messages, we want to keep on forever if possible
```
akramtexas: - Trying to reconcile this with the idea expressed earlier (during a recent `CloudState Contributors’ Call Meeting`) regarding the following two options: (1) Produce, without _consuming_, or (2) Produce without _side-effects_. But I totally agree that `compaction`-related concerns are relatively minor at this time 👍

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
```diff
@@ -0,0 +1,330 @@
...
+    require(upstreamAckDeadline >= 10.seconds,
+            s"upstream-ack-deadline must be at least 10 seconds but was ${upstreamAckDeadline}")
+    require(upstreamAckDeadline <= 600.seconds,
+            s"upstream-ack-deadline must be at most 600 seconds but was ${upstreamAckDeadline}")
+    require(downstreamBatchSize > 0, s"downstream-batch-size must be at least 1 but was ${downstreamBatchSize}")
+  }
+
+  validate()
+
+  // Create the gRPC clients used to communicate with Google Pubsub
```
akramtexas: - It may be helpful to elaborate this (already-useful_ comment to _also_ mention that while the data will be routinely coming in over an RPC channel, it should not make any material difference whether HTTP, or any (other) transport (semantic) is used as well.

proxy/core/src/main/scala/io/cloudstate/proxy/protobuf/Utilities.scala
```diff
@@ -0,0 +1,65 @@
...
+import scala.collection.JavaConverters._
+
+object Types {
+  final val AnyTypeUrlHostName = "type.googleapis.com/"
+}
+
+object Options {
+
+  /**
+   * ScalaPB doesn't do this conversion for us unfortunately.
```
akramtexas: - Is it the case that `ScalaPB`-generated code includes an asynchronous client as well as base traits which can be extended to implement server-side? I would think, though, that such being the case would not affect us (in `serverless`, right?)

proxy/proxy-tests/src/test/scala/io/cloudstate/proxy/eventing/GooglePubsubSpec.scala
```diff
@@ -0,0 +1,133 @@
...
+    userFunction = Process(Seq("node", "index.js"), crdtExampleDir).run()
+
+    node = CloudStateProxyMain.start(
+      ConfigFactory.load("dev-mode.conf").withFallback(ConfigFactory.defaultReference()).resolve()
+    )
+
+    akka.serialization.JavaSerializer.currentSystem.withValue(testSystem.asInstanceOf[ExtendedActorSystem]) {
+      import akka.pattern.ask
+      implicit val timeout = Timeout(1.seconds)
+      val manager = node.actorSelection(node / "server-manager-supervisor" / "server-manager")
```
akramtexas: - Would this result, when invoked (downstream), in the restart of the subordinate, along with clearing out its accumulated state?

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/EventingManager.scala
viktorklang: @akramtexas No, that part of things is the generated IDs produced by the publishing to Pubsub (or any other messages which may be produced as the result of publication.

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
viktorklang: @akramtexas This file is GooglePubsubEventing.scala, which is why it specifically mentions this detail.

proxy/core/src/main/scala/io/cloudstate/proxy/protobuf/Utilities.scala
viktorklang: We use this since the binary proto definitions we get are not compiled—they are received at runtime.

proxy/proxy-tests/src/test/scala/io/cloudstate/proxy/eventing/GooglePubsubSpec.scala
viktorklang: This is a test case, and the only thing we do here is to make sure that the server eventually becomes ready to accept requests.

viktorklang:
Rebased on current master

viktorklang:
Merging this so we can iterate on the feature.

Issue #51, opened 2019-07-18
How to do KVS


viktorklang:
Having support for Key-Value style state management would be interesting,
it could either be implemented on top of CRUD #50 by the Entity being a Map.

This also impacts the user-facing API as they would not have to deal with anything but the inbound commands (not events), by adding/modifying/removing the values in the Map.

kwehden:
We should consider doing this on Postgres, to demonstrate that we work with other persistence stores outside of Cassandra.  This I think is a perceived bias that we've seen in some feedback.

viktorklang:
@kwehden Agreed—I'm a big fan of Postgres as an RDBMS

andyczerwonka:
@kwehden & @viktorklang using native hstore and/or JSONB within Postgres may make sense for this use case

PR #143, opened 2019-11-14
Some food-for-thought for supporting Key-Value API, this shows we can…

viktorklang:
… implement it as a pure lang-support alternative.

Related to https://github.com/cloudstateio/cloudstate/issues/51

### DIFF
--- a/java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
+++ b/java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
@@ -1,0 +1,200 @@
+package io.cloudstate.javasupport;
+
+import java.util.Optional;
+import java.lang.Comparable;
+import java.util.function.Function;
+import static java.util.Objects.requireNonNull;
+import static java.util.Collections.singletonMap;
+import akka.util.ByteString;
+
+import io.cloudstate.keyvalue.KeyValue.KVEntityOrBuilder;
+import io.cloudstate.keyvalue.KeyValue.KVEntity;
+import io.cloudstate.keyvalue.KeyValue.KVModification;
+import io.cloudstate.keyvalue.KeyValue.KVModificationOrBuilder;
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+
+public final class KeyValue {
+  public static final class Key<T> implements Comparable<Key<?>> {
+    private final String _name;
+    private final Function<ByteString, T> _reader;
+    private final Function<T, ByteString> _writer;
+
+    Key(
+        final String name,
+        final Function<ByteString, T> reader,
+        final Function<T, ByteString> writer) {
+      this._name = name;
+      this._reader = reader;
+      this._writer = writer;
+    }
+
+    public final String name() {
+      return this._name;
+    }
+
+    public final Function<ByteString, T> reader() {
+      return this._reader;
+    }
+
+    public final Function<T, ByteString> writer() {
+      return this._writer;
+    }
+
+    @Override
+    public final boolean equals(final Object that) {
+      if (this == that) return true;
+      else if (that instanceof Key) return ((Key<T>) that).name().equals(this.name());
+      else return false;
+    }
+
+    @Override
+    public final int hashCode() {
+      return 403 + name().hashCode();
+    }
+
+    @Override
+    public final int compareTo(final Key<?> other) {
+      return name().compareTo(other.name());
+    }
+  }
+
+  public static final <T> Key<T> keyOf(
+      final String name,
+      final Function<ByteString, T> reader,
+      final Function<T, ByteString> writer) {
+    return new Key<T>(
+        requireNonNull(name, "Key name cannot be null"),
+        requireNonNull(reader, "Key reader cannot be null"),
+        requireNonNull(writer, "Key writer cannot be null"));
+  }
+
+  public static final class Map {
+    /*
+      TODO document invariants
+    */
+    private final java.util.Map<String, ByteString> unparsed;
+    private final java.util.Map<Key<?>, Object> updated = new java.util.TreeMap<Key<?>, Object>();
+    private final java.util.Set<String> removed = new java.util.TreeSet<String>();
+
+    Map(final java.util.Map<String, ByteString> initial) {
+      this.unparsed = requireNonNull(initial, "Map initial values must not be null");
+    }
+
+    public final <T> Optional<T> get(final Key<T> key) {
+      final T value = (T) updated.get(key);
+      if (value != null) {
+        return Optional.of(value);
+      } else {
+        final ByteString bytes = unparsed.get(key.name());
+        if (bytes != null) {
+          final T parsed = key.reader().apply(bytes);
+          requireNonNull(parsed, "Key reader not allowed to read `null`");
+          updated.put((Key<Object>) key, parsed);
+          return Optional.of(parsed);
+        } else {
+          return Optional.empty();
+        }
+      }
+    }
+
+    public final <T> void set(final Key<T> key, final T value) {
+      requireNonNull(key, "Map key must not be null");
+      requireNonNull(value, "Map value must not be null");
+      updated.put(key, value);
+      unparsed.remove(key.name());
+      removed.remove(key.name());
+    }
+
+    public final <T> boolean remove(final Key<T> key) {
+      requireNonNull(key, "Map key must not be null");
+      if (!removed.contains(key.name())
+          && (updated.remove(key) != null
+              | unparsed.remove(key.name()) != null)) { // Yes, you read that strict-or right
+        removed.add(key.name());
+        return true;
+      } else {
+        return false;
+      }
+    }
+
+    final KVEntityOrBuilder toProto() {
+      final KVEntity.Builder builder = KVEntity.newBuilder();
+      unparsed.forEach(
+          (k, v) -> {
+            builder.putEntries(k, com.google.protobuf.ByteString.copyFrom(v.asByteBuffer()));
+          });
+      updated.forEach(
+          (k, v) -> {
+            // Skip as this item is only a parsed un-changed item, and we already added those in the
+            // previous step
+            if (!unparsed.containsKey(k.name())) {
+              builder.putEntries(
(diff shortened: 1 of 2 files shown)


java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
```diff
@@ -0,0 +1,200 @@
...
+      } else {
+        return false;
+      }
+    }
+
+    final KVEntityOrBuilder toProto() {
+      final KVEntity.Builder builder = KVEntity.newBuilder();
+      unparsed.forEach(
+          (k, v) -> {
+            builder.putEntries(k, com.google.protobuf.ByteString.copyFrom(v.asByteBuffer()));
```
marcellanz: @viktorklang 
This is the serialization for the user choosen for his values in the KVS, right?
Does this conflict or is it compatible with the [stable serialization rules by Cloudstate](https://cloudstate.io/docs/user/lang/java/serialization.html)? Is this relevant here too? I think about how, if done, a separate/different user lib, language wise, would parse the values back is they're are marshalled as ByteStrings here.

viktorklang:
>
>
>  @viktorklang <https://github.com/viktorklang>
> This is the serialization for the user choosen for his values in the KVS,
> right?
> Does this conflict or is it compatible with the stable serialization
> rules by Cloudstate
> <https://cloudstate.io/docs/user/lang/java/serialization.html>? Is this
> relevant here too? I think about how, if done, a separate/different user
> lib, language wise, would parse the values back is they're are marshalled
> as ByteStrings here.
>


It’s just food-for-thought, but you’re right—for this to be correct we’d
have to use the same serialization scheme as for the rest. I just did it
like this for illustrative purposes 🙂

> --
Cheers,
√

marcellanz:
> It’s just food-for-thought, but you’re right—for this to be correct we’d have to use the same serialization scheme as for the rest. I just did it like this for illustrative purposes 🙂
> --
> Cheers, √

Cool stuff 👍

viktorklang:
Thought: Since this is more of a pattern than a distinct different storage mode, should we make this a documentation feature only?

java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
viktorklang: Yeah, this needs work. I am not sure how to integrate it though—I'd need to go through AnySupport I think.

marcellanz: Could be a good example how to use AnySupport.

marcellanz:
> Thought: Since this is more of a pattern than a distinct different storage mode, should we make this a documentation feature only?

In #144 @jroper argues, that the user should not have to know that any of the CRDTs he uses is a CRDT, by using one. Similarly a user also should not have to know a "Cloudstate KVS" he uses is an event sourced entity underneath?

java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
viktorklang: Indeed. I'll have a look at what that could look like.

java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
```diff
@@ -0,0 +1,243 @@
...
+
+  // We'll most likely want to add a KeyValueEntity annotation instead
+  public static class YourClass extends KeyValueEntity {
+    private static final Key<String> name =
+        keyOf("name", ByteString::utf8String, ByteString::fromString);
+
+    // We'll most likely want to add a specific CommandContext for KeyValueEntity
+    // so you'd instead of state().set(…) would do ctx.setState(…), and then it'd be automatically
+    // flushing ctx.emit(state.toProtoModification()) if there are any changes
+    @CommandHandler
```
viktorklang: @jroper What do you think about the comments above, James? Do you think this is something we could put together rather quickly?

jboner: I like that idea.

viktorklang: @jroper ping :)

sleipnir: @viktorklang Any progress here? This seems to be stopped for a long time

viktorklang: @sleipnir No, unfortunately not. I guess the question is: Should we go for an API-only approach or support KV natively in the proxy as a different way of storing data.

sleipnir:
Hi @viktorklang 
I am suspicious to answer because, I almost always end up voting for the proxy. But I think delegating to the proxy is the most elegant and transparent approach. Even taking advantage of persistence with Cassandra, Postgres, or key value native applications like Redis, Riak, Hazelcast, Infinispan, or anything like that. It's the kind of thing that would be worth exposing user interfaces to implement its own persistence mechanism (managed by the proxy of course)

java-support/src/main/java/io/cloudstate/javasupport/KeyValue.java
sleipnir: Hi @viktorklang
I am suspicious to answer because, I almost always end up voting for the proxy. But I think delegating to the proxy is the most elegant and transparent approach. Even taking advantage of persistence with Cassandra, Postgres, or key value native applications like Redis, Riak, Hazelcast, Infinispan, or anything like that. It's the kind of thing that would be worth exposing user interfaces to implement its own persistence mechanism (managed by the proxy of course)

viktorklang: Hi @sleipnir,
I'm inclined to agree—I think it is much more valuable for the proxy to understand the specific state model and be able to better accommodate for its access patterns and optimization opportunities.

marcellanz: @viktorklang @sleipnir do we have consensus how we would go forward with this?
@pvlugter @jroper do you have this on your roadmap?

sleipnir: I think someone is going to have to get their hands dirty here and try to create something to start interacting better. I haven't seen any recent progress here @marcellanz . @ralphlaude  has an implementation proposal but I don't know how much he has evolved in that.

ralphlaude: @sleipnir i agree with you, someone have to get their hands dirty here :). I have thinking a bit about the a possible implementation for the native support and also the grpc protocol. But i didn't get that far. I am looking forward to get more involved here when other stuff going on are done :).

@marcellanz very good question and I am not sure there is a consensus here. If there is a consensus, I think the first step here would be to define the grpc protocol.

pvlugter: The CRUD support will provide the underlying native KV support discussed in this thread, and @ralphlaude is making progress there. (Side note: not sure if "CRUD" entities is the best name, maybe we can improve the naming there too.) Until that's complete, language supports could also experiment with including a `setState` or `setValue` style API directly on event sourced entities, setting a snapshot value, if we want to try this out earlier.

Also agree that it would be good to try out some use cases, to see how useful a KV API like in this PR might be, since this would need to be added to all the language supports. This currently gives a more generic Map-like API in place of defining protobuf message objects directly (which can also be map-like as needed). The shopping cart samples also use a map for state — would that example be improved with a generic key-value API, compared with the types it uses now for event sourcing? In these kinds of examples, would just setting a map as the value (and using CRUD support) already be clean enough?

Entity services themselves provide a KV-style store, keyed by entity id, and can be backed by event sourced, or the upcoming CRUD for a simple set-value approach, or using CRDTs for distributed, in-memory, eventually consistent storage. Within an entity the state can be a map (be key-value-based) like in this PR.

And then there's the concept of a KV store that's available across Entity or Action instances — like the Cloudflare Workers KV, which is an eventually-consistent LWW (last writer wins) key-value store distributed across the worker nodes. For Cloudstate, the pattern that we've demonstrated that's similar to this is having multiple entity types. Like the "hot items" CRDT entity used alongside the shopping cart. We have forwarding and side effects to combine entities. Maybe we should look at supporting an "ask pattern", an async get on another entity type, where values can be fetched from other entity types and then continue processing with the responses.

pvlugter: @jroper and I discussed this further. Some things we discussed:

We don't currently see clear use cases for a key-value API like this in the language supports. Especially when compared with simply using an object or map as the value (with CRUD or event sourcing support). For persisting deltas/changes only, there's the event sourcing already, and having an additional KV API increases the surface area that language supports need to implement.

For CRUD support naming, we're suggesting to refer to this as a value-based state model, and it could be the default state model for entities. From the user perspective, we could simply call these `Entities`. And then there are `Event Sourced Entities` and `Replicated Entities` (CRDTs). Within code or discussions, we could use value entities or value-based entities or persisted value entities as needed. And for services, entities themselves provide key-value storage (with the entity key and its value), which can be backed by simple key-value storage, event sourcing, or in-memory replication.

I think the main thing to do here is to complete the basic entities with value-based state model (CRUD support). In terms of roadmap, we also discussed other forward-looking things, like replicated and active-active event-sourcing, durable actions (as in long-lasting, that can fetch asynchronously, implement workflows), durable CRDTs, and product CRDTs.

So suggesting that we close this PR and put the idea of KV APIs in language supports on hold, and focus on finishing the value entities (CRUD support) and making those the default starting point. Interested to hear more thoughts on this, and it can be a topic for the next contributors' call.

jboner: I agree that this is the best path forward.

viktorklang: Having a good "CRUD" solution might obviate the need for a KV solution, especially if the proxy can diff.

pvlugter:
Closing this, as discussed.

PR #153, opened 2019-11-26
Implements support for automatically creating a zip archive of the pr…

viktorklang:
…otocols on package

Related to https://github.com/cloudstateio/cloudstate/issues/118
This would mean that we could have a checked in zip-archive which lang supports could fetch over HTTP and unzip.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -99,18 +99,58 @@ headerSources in Compile ++= {
 lazy val root = (project in file("."))
 // Don't forget to add your sbt module here!
 // A missing module here can lead to failing Travis test results
-  .aggregate(`proxy-core`,
-             `proxy-cassandra`,
-             `proxy-postgres`,
-             `proxy-tests`,
-             `java-support`,
-             `scala-support`,
-             `java-shopping-cart`,
-             `akka-client`,
-             operator,
-             `tck`,
-             docs)
+  .aggregate(
+    `protocols`,
+    `proxy-core`,
+    `proxy-cassandra`,
+    `proxy-postgres`,
+    `proxy-tests`,
+    `java-support`,
+    `scala-support`,
+    `java-shopping-cart`,
+    `akka-client`,
+    operator,
+    `tck`,
+    docs
+  )
   .settings(common)
+
+val cloudstateProtocolsName = "cloudstate-protocols"
+val cloudstateTCKProtocolsName = "cloudstate-tck-protocols"
+
+lazy val protocols = (project in file("protocols"))
+  .settings(
+    name := "protocols",
+    publish / skip := true,
+    packageBin in Compile := {
+      val base = baseDirectory.value
+      val cloudstateProtos = base / s"$cloudstateProtocolsName.zip"
+      val cloudstateTCKProtos = base / s"$cloudstateTCKProtocolsName.zip"
+
+      def archiveStructure(topDirName: String, files: PathFinder): Seq[(File, String)] =
+        files pair Path.relativeTo(base) map {
+          case (f, s) => (f, s"$topDirName${File.separator}$s")
+        }
+
+      // Common Language Support Proto Dependencies
+      IO.zip(
+        archiveStructure(cloudstateProtocolsName,
+                         (base / "frontend" ** "*.proto" +++
+                         base / "protocol" ** "*.proto" +++
+                         base / "proxy" ** "*.proto")),
+        cloudstateProtos
+      )
+
+      // Common TCK Language Support Proto Dependencies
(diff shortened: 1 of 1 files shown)


viktorklang:
Ping @marcellanz

marcellanz:
> Ping @marcellanz

@viktorklang tested and works for me.

viktorklang:
@marcellanz Cool, thanks for verifying! Should we include this in the developer documentation as well perhaps?

marcellanz:
> @marcellanz Cool, thanks for verifying! Should we include this in the developer documentation as well perhaps?

@viktorklang sure :-)

marcellanz:
@viktorklang examples was left out by intention right? Language support would need it to implement the TCK. "examples" are not part of the spec, but if the consumer if the zip is language support or alternative proxy implementers, the examples are needed by them. WDYT?

viktorklang:
@marcellanz Yeah, I kept the shopping-cart protos outside of it. Do we want to have an extra zip-file for the shopping-cart protos?

marcellanz:
@viktorklang I don't know, really. And name it tck-protos?

viktorklang:
@marcellanz That's a great suggestion, we might have more of those later anyway. I'll amend this PR.

viktorklang:
@marcellanz Check the latest commit(s)

marcellanz:
@viktorklang 

works... should it get cleaned by `sbt:cloudstate> clean`?

Screenshot 2019-11-28 at 10 31 58

(digged into sbt documentation, let me know if it should get cleaned and I'll find out how, having to learn some sbt I think)

viktorklang:
@marcellanz I'll add cleaning

viktorklang:
@marcellanz Added the files to clean.

build.sbt
```diff
@@ -99,19 +99,56 @@ headerSources in Compile ++= {
...
+        commonProtos
+      )
+
+      // Common TCK Language Support Proto Dependencies
+      IO.zip(archiveStructure("cloudstate-tck-protocols", base / "example" ** "*.proto"), tckProtos)
+
+      commonProtos
+    },
+    cleanFiles ++= Seq(
+        baseDirectory.value / (name.value + "-" + version.value + ".zip"),
```
marcellanz: @viktorklang this would miss the version if the branch is updated, ok for you?

viktorklang: @marcellanz I opted for being careful deleting files outside of the target directory. Perhaps we should not include the version number in the archive name at all?

marcellanz: >Perhaps we should not include the version number in the archive name at all?

@viktorklang That makes sense I think. One would get the zip like so perhaps:
```
curl -O https://raw.githubusercontent.com/cloudstateio/cloudstate/v0.5.0/protocols/cloudstate-protos.zip
```
where its clear which version one gets. Same for every other tag or branch.

viktorklang: :+1: Will amend.

viktorklang: @marcellanz Done! See latest commit

marcellanz: fetched and tested on my side, works! \o/

viktorklang: @marcellanz Awesome! :-)

ihostage:
Hi, folks 👋 
@marcellanz When this PR will be merged, will you migrate a `go-support` to using these artifacts?

marcellanz:
> Hi, folks 👋
> @marcellanz When this PR will be merged, will you migrate a `go-support` to using these artifacts?

@ihostage Its not a high priority yet, but yes I think user support libraries should get the protobuf files through these artefacts. Currently `go-support`has them, together with the `*.pb.go` files, distributed with a release. Do you have something in mind?

ihostage:
> Do you have something in mind?
@marcellanz Nothing concrete 😄 
Just I wait for these artifacts and try to find time to play with another language support 😄
But the end of the year is a hard period and find time is very difficult. 😞

marcellanz:
> > Do you have something in mind?
> 
> @marcellanz Nothing concrete 😄
> Just I wait for these artifacts and try to find time to play with another language support 😄
> But the end of the year is a hard period and find time is very difficult. 😞

@ihostage cool! Yeah time... I know, one can squeeze for a few weeks but then one have to pay tribute and compensate. Which language do you have in mind?

viktorklang:
@marcellanz Merged! Also, added notes in https://github.com/cloudstateio/cloudstate/blob/master/docs/src/main/paradox/developer/language-support/index.md#creating-language-support-libraries

marcellanz:
@viktorklang awesome, thanks!

PR #154, opened 2019-11-27
Work-in-progress on adding support for StatelessFunctions in the proxy

viktorklang:
Opening this PR to gather some input and feedback (and contributors!), as this will require implementing support in each language library after the feature is deemed ready on the proxy (for integration that is).


viktorklang:
@jroper Merging this on the proxy side so progress on language support can be begun.

PR #158, opened 2019-12-03
Experimental new implementation of Eventing which allows for publishi…

viktorklang:
…ng the results of any called endpoint no matter how it was invoked

Just a draft PR, left to do is to devise a good test suite for it, and bake that into the TCK.
Will update this if the design is deemed sound.


viktorklang:
Merging … next step is to add verification to the TCK

Issue #159, opened 2019-12-04
500 response when sending malformed JSON


jsravn:
I deployed:
```
apiVersion: cloudstate.io/v1alpha1
kind: StatefulService
metadata:
  name: shopping-cart
spec:
  containers:
  - image: cloudstateio/samples-js-shopping-cart:latest
    name: "shopping-cart"
```

Then curl with malformed JSON:
```
curl -v -H "content-type: application/json" -X POST --data '{"product_id":"1", "name":"engine", "quantity":10,}' http://172.17.0.2:30520/cart/james/items/add
```

(Note trailing comma on `"quantity":10,` but any malformed JSON seems to do this)

Then I get back a 500 error rather than 400:
```
2019-12-04 14:20:16.194 ERROR akka.actor.ActorSystemImpl - Internal server error, sending 500 response                                                                                                                                                                                                                        
com.google.protobuf.InvalidProtocolBufferException: com.google.gson.stream.MalformedJsonException: Expected name at line 5 column 2 path $.quantity 
```

viktorklang:
Fixed!

PR #160, opened 2019-12-04
Treats request processing failures as Bad Request instead of Internal…

viktorklang:
… Server Error


viktorklang:
Addresses https://github.com/cloudstateio/cloudstate/issues/159

PR #173, opened 2019-12-20
Explicit Metaspace configuration in `.sbtopts`

ChristopherAndre:
Suggestion to make metaspace an explicit parameter as it otherwise would cause `sbt clean test` to fail on certain systems.


viktorklang:
Thanks @ChristopherAndre!

PR #174, opened 2019-12-28
Added further information about the content of the `frontend` folder

ChristopherAndre:


docs/src/main/paradox/developer/language-support/index.md
```diff
@@ -23,18 +23,58 @@ To obtain the necessary Cloudstate Protobuf descriptors which needs to be implem
...
 
 - `example` is for implementing the example application, which is also used by the TCK for third-party language (and proxy implementation) validation.
-- `frontend` is what is used by developers of services when they define their proto interfaces.
-- `protocol` is the protocol between the proxy and what we call a Language Support, i.e. a bridge library which speaks with the proxy and exposes a native API for some programming language.
-- `proxy` contains the protocols which the proxy itself speaks with the outside world.
-
-When implementing a Support Library, the implementation can be verified using the TCK. In order to run the TCK, one must create an implementation of the Shopping Cart application using the newly created Support Library which is to be verified.
+- `frontend` is what is used by developers of services when they define their proto interfaces. The `cloudstate` subfolder in particular contains the definition of two protobuf options:
+    
+    - `.cloudstate.entity_key`: used to indicate the String to be used as an entity key in a given protobuf object.
```
viktorklang:
```suggestion
    - `.cloudstate.entity_key`: used to indicate the String to be used as an entity key in a given protobuf object. Multiple fields can be annotated to form a composite key.
```

ChristopherAndre: @viktorklang The `entity_key` on the proxy side is going to be a simple String though, right? If so, how are multiple annotated fields combined together to give us the same `entity_key` on the proxy side? Is this behaviour defined somewhere?

viktorklang: That should be in the spec at some point, right now it's only in code, I'd send a link but I'm running to a meeting right now. Let me know if you don't find it and I'll have a look later!

viktorklang:
Thanks @ChristopherAndre!

Issue #178, opened 2020-01-10
Emojis in user documentation do not render properly in the browser


marcellanz:
**_What did you see?_**

https://cloudstate.io/docs/user/lang/index.html#3rd-party-libraries
<img width="519" alt="Screenshot 2020-01-10 at 11 42 38" src="https://user-images.githubusercontent.com/12616445/72147211-4f822500-339e-11ea-8088-172eb805f1df.png">

**_What did you expect to see?_**
more emojions

**_Additional Information_**
#170 might have introduced that.

michaelpnash:
The link "source code for this page here" *does* show all the icons, it seems, interestingly.

marcellanz:
@michaelpnash It may be the markdown parser not capable of, but I'm not an "emoji" expert :-)

sleipnir:
I suggest removing the emojis since they were inherited from another repository and may not make sense here.
We can simply write (supported, not supported, partially supported, unstable, etc ...)

sleipnir:
@ChristopherAndre what do you think?

ChristopherAndre:
@sleipnir : You're right, let's use single letters instead! What about **"S"** for "Supported", **"N"** for "Not Supported", **"P"** for "Partial support" and **"?"** for "Support status unknown" ?

viktorklang:
I'd rather have it in text instead of a letter, that will translate easier using Google Translate etc.

ChristopherAndre:
@viktorklang There would still be a legend of what these letters mean.

ChristopherAndre:
I have created a PR to address that: basically, I converted the emoji reference to their unicode representation. Hopefully, that should solve the problem!
Let me know if you prefer to have full text instead!

sleipnir:
> I have created a PR to address that: basically, I converted the emoji reference to their unicode representation. Hopefully, that should solve the problem!
> Let me know if you prefer to have full text instead!

Cool!

ChristopherAndre:
PR has been merged.
@marcellanz please test on your browser(s) and let us know if it works for you!

jroper:
Thanks @ChristopherAndre, yes, the colon - emoji name - colon syntax is not markdown, so there's no reason to expect it to work, it works in GitHub because GitHub uses their own flavour of markdown called GitHub Flavoured Markdown where they've added emoji support.

[...]

PR #181, opened 2020-01-15
Attempt to solve issue 178

ChristopherAndre:
Attempt to address issue #178
This seems to work locally but would probably need further tests on other machines.


viktorklang:
Thanks @ChristopherAndre!

PR #186, opened 2020-01-26
[fix] this might be a typo.

marcellanz:
@coreyauger I'm reading the node stateless code and found a typo I think (`tservice vs. service`) on the way and attached a fix for it.

https://github.com/cloudstateio/cloudstate/blob/a70dece2cd2a90397db09fc31fd4ff7f879fd61f/node-support/src/stateless-support.js#L135


viktorklang:
JS O_o

PR #189, opened 2020-02-04
circleci config for automated integrated tests with native images

srikanthops:
CIrcleCI config for 
a) building native image
b) it test
c) it test with native image 

Node setup: used apt to install latest node, as found several issues with  bash/nvm based installation.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,175 @@
+version: 2.1
+# CircleCi Build Config for CloudState
+commands:
+  setup_sbt:
+    description: "Set up SBT"
+    parameters:
+      version:
+        type: string
+        default: "1.3.3"
+    steps:
+      - run:
+          name: Install SBT
+          command: |
+            curl -L -o sbt-<< parameters.version >>.deb https://dl.bintray.com/sbt/debian/sbt-<< parameters.version >>.deb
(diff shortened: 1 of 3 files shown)


viktorklang:
Nice work, @srikanthops! @jroper, would you be able to review this?

jroper:
It appears the circle ci setup is currently skipping running the tests, is that intentional?

Is the plan that we build the native image in one job, and then test it in another? I guess that requires sharing the artifacts using the build workspace, but I can't see that being done anywhere.

What do we need to do to enable CircleCI on this repo? Will we have to create a paid account? I think we should enable it before merging this PR, so we can test that this PR passes.

srikanthops:
Thanks James for the initial comments

>> It appears the circleCi setup is currently skipping running the tests, is that intentional?
Enabled tests

>>Is the plan that we build the native image in one job, and then test it in another? I guess that requires 
>> sharing the artifacts using the build workspace, but I can't see that being done anywhere.

No at present we are not sharing  artifacts from build, I believe the tests doesn't even need all the artifacts..
The build takes lot of time compared to the tests

What do we need to do to enable CircleCI on this repo? 
We should be able to enable this on CircleCI, since this repo is open sourced, it can be tested for free on circleCI.

>>  I think we should enable it before merging this PR, so we can test that this PR passes.
That would be the right thing to do, Let me see If I can do that

PR #190, opened 2020-02-09
Fix <dependency> tag

sleipnir:
Replace the incorrect depependency with the correct dependency tag


viktorklang:
Thanks @sleipnir!

PR #192, opened 2020-02-10
Settings ServiceInfo in order to enable the reuse of java-support in other JVM languages

sleipnir:
Enables us to customize the information reported in ServiceInfo

### DIFF
--- a/java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
+++ b/java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
@@ -25,11 +25,13 @@ import io.cloudstate.javasupport.{BuildInfo, StatefulService}
 
 class EntityDiscoveryImpl(system: ActorSystem, services: Map[String, StatefulService]) extends EntityDiscovery {
 
+  private val config = system.settings.config.getConfig("cloudstate")
+
   private val serviceInfo = ServiceInfo(
     serviceRuntime = sys.props.getOrElse("java.runtime.name", "") + " " + sys.props.getOrElse("java.runtime.version",
                                                                                               ""),
-    supportLibraryName = sys.env.getOrElse("SUPPORT_LIBRARY_NAME", BuildInfo.name),
-    supportLibraryVersion = sys.env.getOrElse("SUPPORT_LIBRARY_VERSION", BuildInfo.version)
+    supportLibraryName = config.getString("library.name"),
+    supportLibraryVersion = config.getString("library.version")
   )
 
   /**
--- a/java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
+++ b/java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
@@ -28,8 +28,8 @@ class EntityDiscoveryImpl(system: ActorSystem, services: Map[String, StatefulSer
   private val serviceInfo = ServiceInfo(
     serviceRuntime = sys.props.getOrElse("java.runtime.name", "") + " " + sys.props.getOrElse("java.runtime.version",
                                                                                               ""),
-    supportLibraryName = BuildInfo.name,
-    supportLibraryVersion = BuildInfo.version
+    supportLibraryName = sys.env.getOrElse("SUPPORT_LIBRARY_NAME", BuildInfo.name),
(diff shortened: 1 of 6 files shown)


java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
```diff
@@ -28,8 +28,8 @@ class EntityDiscoveryImpl(system: ActorSystem, services: Map[String, StatefulSer
   private val serviceInfo = ServiceInfo(
     serviceRuntime = sys.props.getOrElse("java.runtime.name", "") + " " + sys.props.getOrElse("java.runtime.version",
                                                                                               ""),
-    supportLibraryName = BuildInfo.name,
-    supportLibraryVersion = BuildInfo.version
+    supportLibraryName = sys.env.getOrElse("SUPPORT_LIBRARY_NAME", BuildInfo.name),
```
viktorklang: Instead of pulling it from the env, I think it makes more sense to pull it from the system.settings.config

sleipnir: Ok, i send commit later

viktorklang: @sleipnir Awesome. Let me know if you need my input!

sleipnir: Ok @viktorklang I sent a new attempt, can you check if you agree?

java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
```diff
@@ -25,11 +25,13 @@ import io.cloudstate.javasupport.{BuildInfo, StatefulService}
...
+  private val config = system.settings.config.getConfig("cloudstate")
+
   private val serviceInfo = ServiceInfo(
     serviceRuntime = sys.props.getOrElse("java.runtime.name", "") + " " + sys.props.getOrElse("java.runtime.version",
                                                                                               ""),
-    supportLibraryName = BuildInfo.name,
-    supportLibraryVersion = BuildInfo.version
+    supportLibraryName = config.getString("library.name"),
+    supportLibraryVersion = config.getString("library.version")
   )
```
viktorklang: :+1:

sleipnir:
@viktorklang approved ?

viktorklang:
Thanks @sleipnir! :)

sleipnir:
It's nothing!

java-support/src/main/scala/io/cloudstate/javasupport/impl/EntityDiscoveryImpl.scala
```diff
@@ -25,11 +25,13 @@ import io.cloudstate.javasupport.{BuildInfo, StatefulService}
...
 
+  private val config = system.settings.config.getConfig("cloudstate")
+
   private val serviceInfo = ServiceInfo(
     serviceRuntime = sys.props.getOrElse("java.runtime.name", "") + " " + sys.props.getOrElse("java.runtime.version",
                                                                                               ""),
-    supportLibraryName = BuildInfo.name,
-    supportLibraryVersion = BuildInfo.version
+    supportLibraryName = config.getString("library.name"),
+    supportLibraryVersion = config.getString("library.version")
```
pvlugter, line 34: Does it matter that this no longer uses the automatically generated current version for cloudstate-java-support, but is instead fixed in config?

sleipnir: @pvlugter  This exists for another support idiom library (like Kotlin) that uses lib java-support as a base and defines a hocon file reference.conf or application.conf and can overwrite the version of the java library with its own information.

pvlugter: Sure, understand the usage. But currently published versions of cloudstate-java-support, like 0.4.7, will report their version as 0.4.3 since it's hardcoded in the config. What we want is for the name and version to be overridable to config, but use the generated build info by default. I'll make an update for this.

pvlugter: Updated in #294

PR #200, opened 2020-02-25
conversion of skin to new branding and colors

rasummer:
@jroper this updates the docs skin with the new logo and colors. 

CSS is now generated from a Stylus file. CD into the CSS directory and run `stylus -c --watch page.styl` to develop locally with the CSS. I wasn't sure where we should document this. I tried adding a readme.md in the CSS directory, but paradox didn't like that. 

cloudstate-docs-skin


viktorklang:
@rasummer @jroper Thanks Richard! :)

PR #204, opened 2020-03-07
Disable graalvm build

srikanthops:
Disabling graalvm build as it requires `resource_class:large` machine.
Observed incompatibilities for scala format version and graalvm version, so use 
`set scalafmtOnCompile := false`


.circleci/config.yml
```diff
@@ -131,7 +131,8 @@ jobs:
           name: graal build
           command: |
             export PATH="~/graal/graalvm-ce-java11-19.3.0/bin:$PATH"
-            sbt "project proxy-core" "set graalVMVersion := None"  graalvm-native-image:packageBin
+            # disabling sclalaformating, observed some incompatibilities between graalvm version and scalafmt version.
```
viktorklang:
```suggestion
            # disabling scalaformating, observed some incompatibilities between graalvm version and scalafmt version.
```

viktorklang:
LGTM after fixing typo. @jroper @edwardcallahan @michaelpnash We need to make sure that we validate the graalvm native-image atleast daily (assuming that there has been any merges since last successful run)

ihostage:
38 commits for changing of some lines 😱 
The hell for the perfectionists, that love the clean git history 😂

Issue #219, opened 2020-03-25
To make java-support more friendly to CDI environments


sleipnir:
To make java-support more friendly to CDI environments like creating a registerEventSourcedEntity and registerCrdtEntity that accepts a class instance instead of Class.

For that I believe that small changes are needed in Cloudstate.java, AnnotationBaseadEventSourcedSupporte.scala, and AnnotationBasedCrdtSupport.scala.

For more information see: https://github.com/sleipnir/spring-boot-cloudstate-starter/issues/2

sleipnir:
More details: https://github.com/sleipnir/spring-boot-cloudstate-starter/issues/2#issuecomment-604468010

sleipnir:
ping @viktorklang @marcellanz

marcellanz:
@sleipnir sorry to not write yet; too much is happening (at the same time) at the moment.

viktorklang:
@sleipnir Apologies, it's been crazy busy this week. Revisit early next? /cc @jroper

sleipnir:
No problem guys, I’m very busy myself with everything that’s going on too. I know it's hard.

sleipnir:
Looking at the code again, I saw that we only need a new EventSourcedEntityFactory and CrdtEntityFactory that accept a factory class that works as an 'extension' so that each time the create method is called, this fabric delegates to the extension so that we can obtain the instance directly from Spring (in this case the bean must be declared with the prototype Scope so that a new clean object instance is always created), or another context injection.
With that I could use the existing method in Cloudstate.java:

```
public CloudState registerEventSourcedEntity (
      EventSourcedEntityFactory factory,
      Descriptors.ServiceDescriptor descriptor,
      String persistenceId,
      int snapshotEvery,
      Descriptors.FileDescriptor ... additionalDescriptors) {
      ......
}
```

To register user entity classes in a simple way.

It is clear that this new factory should be able to inject the entityId and the Context in a way other than via the constructor. Probably via setter or class attribute.

The question is whether this should be inside java-support or outside. If the class were not really complex as it would be, it would be simple to implement it in the scope of Spring, but as it has a lot of knowledge of the internal structures it may not be so simple to replicate that outside the library......

viktorklang:
@sleipnir TBH It hink having a separate `java-spring-support` module makes more sense. I assume that there'll be others.

sleipnir:
Agree. Is the same is the case with the AKKA Extensions module

[...]

PR #227, opened 2020-04-01
WIP DI Support

sleipnir:
Hi @viktorklang 

Here I demonstrate how we could implement the functionality described in the issue https://github.com/cloudstateio/cloudstate/issues/219
  It is a proof of concept, I was more concerned with the parts and not with their design so if there is anything strange in the Scala code feel free to suggest a code to replace.

You can follow the client side of this API in the repository https://github.com/sleipnir/spring-boot-cloudstate-starter/tree/wip-di-support

I managed to run all the expected features of object injection created via spring with this code that I am proposing, that is, all Spring features related to dependency injection are compatible with this proposal.
Look in the examples folder of the aforementioned repository.


java-support/src/main/java/io/cloudstate/javasupport/EntitySupportFactory.java
```diff
@@ -0,0 +1,18 @@
...
+public interface EntitySupportFactory {
+  /**
+   * The reference to the Entity.
+   *
+   * @return The reference to the Entity.
+   */
+  Object create(Context context, String enittyId);
+
+  /**
+   * Return Class of Etity.
```
viktorklang, line 13:
```suggestion
   * Return Class of Entity.
```

java-support/src/main/java/io/cloudstate/javasupport/EntitySupportFactory.java
```diff
@@ -0,0 +1,18 @@
...
+   *
+   * @return The Class of Etity.
```
viktorklang, line 15:
```suggestion
   * @return The Class of Entity.
```

java-support/src/main/java/io/cloudstate/javasupport/EntitySupportFactory.java
```diff
@@ -0,0 +1,18 @@
+package io.cloudstate.javasupport;
+
+/** Auxiliar interface for support DI environments */
+public interface EntitySupportFactory {
+  /**
+   * The reference to the Entity.
+   *
+   * @return The reference to the Entity.
+   */
+  Object create(Context context, String enittyId);
```
viktorklang, line 10: Can't the return type be more specific than `Object`?

sleipnir: Hummm I don't know. Unless user entities implement some type of interface, I think it would not be possible to determine the correct type.

java-support/src/main/scala/io/cloudstate/javasupport/impl/crdt/AnnotationBasedCrdtExtensionSupport.scala
```diff
@@ -0,0 +1,118 @@
...
+         anySupport,
+         anySupport.resolveServiceDescriptor(serviceDescriptor))
+
+  private val (commandHandlers, streamedCommandHandlers) = {
+    val allMethods = ReflectionHelper.getAllDeclaredMethods(entitySupportFactory.typeClass())
+
+    ReflectionHelper.validateNoBadMethods(allMethods, classOf[CrdtEntity], Set(classOf[CommandHandler]))
+    val handlers = allMethods
+      .filter(_.getAnnotation(classOf[CommandHandler]) != null)
+      .map { method =>
```
viktorklang, line 40:
```suggestion
      .collect { case method if method.getAnnotation(classOf[CommandHandler]) != null =>
```

java-support/src/main/scala/io/cloudstate/javasupport/impl/crdt/AnnotationBasedCrdtExtensionSupport.scala
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    private def unwrap[T](block: => T): T =
+      try {
+        block
+      } catch {
+        case ite: InvocationTargetException if ite.getCause != null =>
+          throw ite.getCause
+      }
+  }
```
viktorklang, line 116: Might be cleaner with

```suggestion
    val andUnwrap: PartialFunction[Throwable, Nothing] = {
        case ite: InvocationTargetException if ite.getCause != null =>
          throw ite.getCause
      }
```

and then instead of:

`unwrap { … }`

you can do:

` try { … } catch andUnwrap`

java-support/src/main/scala/io/cloudstate/javasupport/impl/crdt/CrdtHelper.scala
```diff
@@ -0,0 +1,148 @@
...
+  MainArgumentParameterHandler,
+  MethodParameter,
+  ParameterHandler
+}
+import io.cloudstate.javasupport.impl.{ReflectionHelper, ResolvedType}
+import io.cloudstate.javasupport.{ServiceCall, ServiceCallFactory}
+
+import scala.reflect.ClassTag
+
+class CrdtHelper {
```
viktorklang, line 20: Should this be a class? Seems like it should be an object?

viktorklang:
@jroper @hseeberger Please have a look at this and give @sleipnir feedback—thanks!

java-support/src/main/scala/io/cloudstate/javasupport/impl/crdt/CrdtHelper.scala
sleipnir: ok fixed

sleipnir:
> @jroper @hseeberger Please have a look at this and give @sleipnir feedback—thanks!

???

hseeberger:
@sleipnir unfortunately I have to concede that I am not yet familiar enough with language support for cloudstate.io to give any valuable feedback at this moment in time.

sleipnir:
> @sleipnir unfortunately I have to concede that I am not yet familiar enough with language support for cloudstate.io to give any valuable feedback at this moment in time.

No problem, but I still appreciate your attention.
However I would like to say that the support for the PR functionality has been commented on a lot by some users in private conversations so I think that would be really cool to have on the platform.

Thank you @hseeberger

sleipnir:
Hello Guys @viktorklang @hseeberger  
I think I found a way for you to validate this PR. You just have to look at it without passing the dependency injection. Think of it just as normal Scala code that receives an entity object factory instead of a Class. Validate that it is possible to verify that the delegation to the factory is correct as it was implemented and only. This will be enough to check whether the code works or not.
The DI part is for those who use the client-side functionality.
Can it be like this ? And remember that I am not an expert on Scala so suggest code corrections if any, it makes my life easier ;)

hseeberger:
This might be interesting for @pvlugter ...

sleipnir:
Awesome @pvlugter   i think its approach is worth testing. I make the necessary changes to the spring starter after you make the changes here.
ok for you like that?
Thanks for the suggestions

pvlugter:
@sleipnir, sure, added the suggested changes in #282.

pvlugter:
Replaced by #282

PR #237, opened 2020-04-14
Disable buffering in native-image call

dwijnand:
The call to `!` runs "runBuffered".

This way you get immediate feedback.


viktorklang:
Thanks @dwijnand!

jroper:
Good fix.

PR #239, opened 2020-04-14
Setup a resource generator to convert HOCON to JSON

dwijnand:


proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
```diff
@@ -0,0 +1,59 @@
...
+  }
+  {
+    name: "io.cloudstate.proxy.autoscaler.KubernetesDeploymentScaler$DeploymentList"
+    allDeclaredFields: true
+    allPublicMethods: true
+    allPublicConstructors: true
+  }
+
+  # Other
+  { name: com.google.protobuf.any.Any }
```
viktorklang, line 58: Missing string literal?

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
```diff
@@ -0,0 +1,59 @@
+[
+  {
+    name: "io.cloudstate.proxy.autoscaler.KubernetesDeploymentScaler$Metadata"
```
viktorklang, line 3: I'm not sure I understand the WHY of this PR, but if we are switching to HOCON, would it not make sense to switch to a less verbose format?

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
```diff
@@ -0,0 +1,59 @@
...
+    allDeclaredFields: true
+    allPublicMethods: true
+    allPublicConstructors: true
+  }
```
viktorklang, line 7: No commas needed?

jroper:
So are you thinking of extending this with transformations? Eg, being able to simply declare a list of classnames that a configuration should apply to, and it will output the json?

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
dwijnand: Nope: https://github.com/lightbend/config/blob/master/HOCON.md#commas

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
dwijnand: https://github.com/lightbend/config/blob/master/HOCON.md#unquoted-strings

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
dwijnand: (answering in the main thread)

dwijnand:
@viktorklang 

> I'm not sure I understand the WHY of this PR, but if we are switching to HOCON, would it not make sense to switch to a less verbose format?

@jroper 

> So are you thinking of extending this with transformations? Eg, being able to simply declare a list of classnames that a configuration should apply to, and it will output the json?

It's unfortunate that simply the HOCON -> JSON translation is still fairly verbose.  But I'm slightly hesitant for us, here, to start inventing our own format.

Here are the options, in my mind:

1. invent our own format
2. switch to programmatic definitions (see `Feature` / `@AutomaticFeature` in https://github.com/oracle/graal/blob/release/graal-vm/20.0/substratevm/REFLECTION.md#manual-configuration) and dedup there in code
3. use Dhall to dedup and refactor (see https://twitter.com/travisbrown/status/1250041413892485121; an example: https://github.com/Gabriel439/dhall-manual/blob/b22f9381256780b32041929187cc67c9857ef5e2/manuscript/02-Refactors.md#simplification), or Jsonnet, but I know next to nothing about Jsonnet

The last one has merits (IMO) but is perhaps too far right now.

I think HOCON already provides some niceties, and I'm hoping to be able to remove large chunks of the config by creating custom `Feature`s - for instance, I'm going to start looking at making all the Akka `*Serializer`s be auto-configured.

So, in a way, I want to keep the remaining verbosity as the impetus for trying to understand the underlying cause.  But if you have specific ideas on how to make it less verbose let me know.

proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
viktorklang, line 58: I think it's mainly the inconsistency—putting names in a String literal seems like the thing to do, rather than wondering why some names are string literals and some not.

dwijnand: Fine.

jroper:
Primary reason I think for doing this is HOCON supports comments, JSON doesn't, it would be soooo nice to be able to put comments, links to issues, etc, in our config.

PR #242, opened 2020-04-15
Auto-register Akka serializer classes

dwijnand:


viktorklang:
Thanks @dwijnand, this is a really promising approach for the remaining config work!

PR #245, opened 2020-04-16
Auto-register Akka actor classes

dwijnand:


viktorklang:
Sweet—awesomely done, @dwijnand!

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaActorRegisterFeature.scala
```diff
@@ -0,0 +1,33 @@
...
+import org.graalvm.nativeimage.hosted.Feature.QueryReachabilityAccess
+import org.graalvm.nativeimage.hosted.{Feature, RuntimeReflection}
+
+import scala.collection.JavaConverters._
+
+@AutomaticFeature
+final class AkkaActorRegisterFeature extends Feature {
+
+  override def duringAnalysis(access: Feature.DuringAnalysisAccess) =
+    reachableSubtypes(access, access.findClassByName("akka.akka.Actor")).foreach { cls =>
```
viktorklang, line 13: Perhaps instead of this, we should instead check invocations of Props-creation, and then add *those* Actors to the reflection configuration?

dwijnand: That would be good, but I don't see any public API in Features for doing that.  We could create our own bytecode traversing and recording, but I'm not sure if that's a good idea.

viktorklang: @dwijnand Would it be possible to do in Substitutions somehow?

dwijnand: I don't see how.

PR #248, opened 2020-04-16
Auto-register Protobuf GeneratedMessage classes

dwijnand:

PR #249, opened 2020-04-16
Dedup identical parts of reflect-config in proxies

dwijnand:
Before:

    $ wc -l proxy/*/src/graal/reflect-config.json proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/*-reflect-config.json*
       1225 proxy/cassandra/src/graal/reflect-config.json
        942 proxy/core/src/graal/reflect-config.json
       1001 proxy/postgres/src/graal/reflect-config.json
         14 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/akka-cluster-reflect-config.json
         77 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/akka-management-reflect-config.json
         63 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
       3322 total

After:

    $ wc -l proxy/*/src/graal/reflect-config.json proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/*-reflect-config.json*
        342 proxy/cassandra/src/graal/reflect-config.json
         73 proxy/core/src/graal/reflect-config.json
        132 proxy/postgres/src/graal/reflect-config.json
         14 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/akka-cluster-reflect-config.json
         77 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/akka-management-reflect-config.json
         63 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/cloudstate-reflect-config.json.conf
       1044 proxy/core/src/main/resources/META-INF/native-image/io.cloudstate/cloudstate-proxy-core/other-reflect-config.json.conf
       1745 total

Builds on #248


proxy/core/src/main/scala/io/cloudstate/proxy/ProtobufGeneratedMessageRegisterFeature.scala
```diff
@@ -0,0 +1,33 @@
...
+    List(
+      "akka.protobuf.GeneratedMessage",
+      "akka.protobuf.GeneratedMessage$Builder",
+      "akka.protobuf.ProtocolMessageEnum",
+      "com.google.protobuf.GeneratedMessageV3",
+      "com.google.protobuf.GeneratedMessageV3$Builder",
+      "com.google.protobuf.ProtocolMessageEnum",
+      "scalapb.GeneratedMessage"
+    ).foreach { className =>
+      val cls = access.findClassByName(className)
```
viktorklang, line 22: Wouldn't we want to error if we cannot find these classes?

viktorklang:
@dwijnand You have some conflicts with master which needs to be resolved—pretty awesome dedup!

proxy/core/src/main/scala/io/cloudstate/proxy/ProtobufGeneratedMessageRegisterFeature.scala
dwijnand: I considered it.  There's two situations where that could happen (as I see it):

1. `scalapb.GeneratedMessage` (as an example) _is_, legitimately, on the classpath, but `access.findClassByName` doesn't see it
2. `scalapb.GeneratedMessage` is not on the classpath

If (2) is the case, then throwing an error is wrong and an annoyance.  If (1) is the case, then would it mean Graal is broken?  Would it take long for us to connect the dots back to the skip here?  🤷  I'm thinking we would, so I went with this impl.

viktorklang: @dwijnand I think the right solution is to do what native-image does—it logs if it cannot find the corresponding class for the configuration.

dwijnand: Yeah, I didn't find a logger anywhere.

If I find one I'll come back and add it here.

PR #252, opened 2020-04-17
Handle StatusRuntimeExecptions (fixes #251)

### LINKED ISSUE #251: gRPC exception handler used by proxy does not respect StatusRuntimeException
The currently used gRPC exception handler, which essentially returns a gRPC `Status` for a given exception, only handles our `CommandException`. But there is also another important exception to deal with: `StatusRuntimeException` from the gRPC API, which wraps an already created `Status`.

By not handling `StatusRuntimeException` we return UNKNOWN to the client instead of the actual underlying status.

hseeberger:


hseeberger:
Fixes #251.

viktorklang:
Thanks @hseeberger!

PR #256, opened 2020-04-21
The Great Unbreakening

viktorklang:
Almost there…

PR #260, opened 2020-04-23
The Unbreakening 2: Double Trouble

dwijnand:
(Supersedes #256.)

### DIFF
--- a/proxy/core/src/main/scala/io/cloudstate/proxy/AkkaActorRegisterFeature.scala
+++ b/proxy/core/src/main/scala/io/cloudstate/proxy/AkkaActorRegisterFeature.scala
@@ -8,26 +8,25 @@ import scala.collection.JavaConverters._
 
 @AutomaticFeature
 final class AkkaActorRegisterFeature extends Feature {
-
+  val akkaActorClass = classOf[akka.actor.Actor]
   override def duringAnalysis(access: Feature.DuringAnalysisAccess) =
-    reachableSubtypes(access, access.findClassByName("akka.akka.Actor")).foreach { cls =>
-      for {
-        context <- getDeclaredField(cls, "context")
-        self <- getDeclaredField(cls, "self")
-      } {
-        RuntimeReflection.register(cls)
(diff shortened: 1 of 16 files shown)


proxy/core/src/main/scala/io/cloudstate/proxy/AkkaActorRegisterFeature.scala
```diff
@@ -5,29 +5,40 @@ import org.graalvm.nativeimage.hosted.Feature.QueryReachabilityAccess
...
-    reachableSubtypes(access, access.findClassByName("akka.akka.Actor")).foreach { cls =>
+  override final def duringAnalysis(access: Feature.DuringAnalysisAccess): Unit = {
+    val akkaActorClass =
+      access.findClassByName(classOf[akka.actor.Actor].getName) // We do this to get compile-time safety of the classes, and allow graalvm to resolve their names
+    if (akkaActorClass != null && access.isReachable(akkaActorClass)) {
       for {
-        context <- getDeclaredField(cls, "context")
-        self <- getDeclaredField(cls, "self")
+        subtype <- access.reachableSubtypes(akkaActorClass).iterator.asScala
+        if subtype != null && !subtype.isInterface && !isAbstract(subtype.getModifiers) && !cache(subtype.getName)
```
viktorklang, line 26: On second thought I might have been too greedy here, there are cases where abstract actor classes should be added, so that their context/self field can be available to their subclasses?

```suggestion
        if subtype != null && !subtype.isInterface && !cache(subtype.getName)
```

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaActorRegisterFeature.scala
```diff
@@ -5,29 +5,40 @@ import org.graalvm.nativeimage.hosted.Feature.QueryReachabilityAccess
 import org.graalvm.nativeimage.hosted.{Feature, RuntimeReflection}
 
 import scala.collection.JavaConverters._
+import java.lang.reflect.Modifier.isAbstract
```
viktorklang, line 8:
```suggestion
```

PR #267, opened 2020-04-24
The Unbreakening V: Revenge of the Git

viktorklang:
…ndle it instead

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -361,10 +361,10 @@ def sharedNativeImageSettings(targetDir: File) = Seq(
   "-H:-PrintUniverse", // if "+" prints out all classes which are included
   "-H:-NativeArchitecture", // if "+" Compiles the native image to customize to the local CPU arch
   "-H:Class=" + "io.cloudstate.proxy.CloudStateProxyMain",
-  //"-J-Xmx10g", // native-image is hungry
-  "--verbose",
+  //"-J-Xmx10g", // native-image is hungry FIXME I don't believe this is properly applied even when --no-server is enabled!
   //"--no-server", // Uncomment to not use the native-image build server, to avoid potential cache problems with builds
   //"--debug-attach=5005", // Debugger makes a ton of sense to use to debug SubstrateVM
+  "--verbose",
   "--report-unsupported-elements-at-runtime", // Hopefully a self-explanatory flag FIXME comment this option out once AffinityPool is gone
   "--enable-url-protocols=http,https",
   "--allow-incomplete-classpath",
@@ -376,14 +376,14 @@ def sharedNativeImageSettings(targetDir: File) = Seq(
     "akka.dispatch.affinity",
     "akka.util",
     "com.google.Protobuf",
+    "com.typesafe.config",
     "java.lang.ref.SoftReference", // https://github.com/oracle/graal/issues/2345
     "java.lang.invoke.MethodHandleImpl" // https://github.com/oracle/graal/issues/2345
   ).mkString("=", ",", ""),
-  "--initialize-at-run-time=" +
+  "-H:ClassInitialization=com.typesafe.config.impl.ConfigImpl$EnvVariablesHolder:rerun",
+  "-H:ClassInitialization=com.typesafe.config.impl.ConfigImpl$SystemPropertiesHolder:rerun",
+  "--initialize-at-run-time" +
   Seq(
-    // We want to delay initialization of these to load the config at runtime
-    "com.typesafe.config.impl.ConfigImpl$EnvVariablesHolder",
-    "com.typesafe.config.impl.ConfigImpl$SystemPropertiesHolder",
     // These are to make up for the lack of shaded configuration for svm/native-image in grpc-netty-shaded
     "com.sun.jndi.dns.DnsClient",
     "io.grpc.netty.shaded.io.netty.handler.codec.http2.Http2CodecUtil",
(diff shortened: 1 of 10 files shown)


sleipnir:
Git doesn’t make it easy Lol

marcellanz:
@viktorklang this feels like… 😅
"Limahl - Never Ending Story - 1984", https://www.youtube.com/watch?v=Gf1WT8VEZxk

viktorklang:
@marcellanz That's funny, because right now the native-image job doesn't terminate on my machine. :D

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaSerializerRegisterFeature.scala
```diff
@@ -11,23 +11,58 @@ import java.util.concurrent.ConcurrentHashMap
...
-    val akkaSerializerClass =
-      access.findClassByName(classOf[akka.serialization.Serializer].getName) // We do this to get compile-time safety of the classes, and allow graalvm to resolve their names
+  private[this] final val registerTheseSerializers = {
+    val c = ConcurrentHashMap.newKeySet[String]
+    c.add("akka.serialization.NullSerializer$")
+    c
+  }
+
+  override final def beforeAnalysis(access: Feature.BeforeAnalysisAccess): Unit = {
+    val config = com.typesafe.config.ConfigFactory.load()
```
viktorklang, line 21: @jroper @dwijnand This is at best an approximation of what the serializers will be, but I really can't be booting `CloudstateMain` here.

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaSerializerRegisterFeature.scala
```diff
@@ -11,23 +11,58 @@ import java.util.concurrent.ConcurrentHashMap
...
+    config
+      .getConfig("akka.actor.serializers")
+      .root
+      .unwrapped
+      .values
+      .iterator
+      .asScala
+      .map(_.toString)
+      .foreach(registerTheseSerializers.add)
+    com.typesafe.config.ConfigFactory.invalidateCaches()
```
viktorklang, line 31: This is an attempt to make sure that we don't lock these values into place at build time, rather than at runtime.

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaSerializerRegisterFeature.scala
```diff
@@ -11,23 +11,58 @@ import java.util.concurrent.ConcurrentHashMap
 @AutomaticFeature
 final class AkkaSerializerRegisterFeature extends Feature {
   private[this] final val cache = ConcurrentHashMap.newKeySet[String]
-  override final def duringAnalysis(access: Feature.DuringAnalysisAccess): Unit = {
-    val akkaSerializerClass =
-      access.findClassByName(classOf[akka.serialization.Serializer].getName) // We do this to get compile-time safety of the classes, and allow graalvm to resolve their names
+  private[this] final val registerTheseSerializers = {
+    val c = ConcurrentHashMap.newKeySet[String]
+    c.add("akka.serialization.NullSerializer$")
```
viktorklang, line 16: This is added by Akka itself, and is thus not in the config.

proxy/core/src/main/scala/io/cloudstate/proxy/AkkaSerializerRegisterFeature.scala
```diff
@@ -11,23 +11,56 @@ import java.util.concurrent.ConcurrentHashMap
 @AutomaticFeature
 final class AkkaSerializerRegisterFeature extends Feature {
   private[this] final val cache = ConcurrentHashMap.newKeySet[String]
+
+  // FIXME THIS IS CURRENTLY DISABLED SINCE IT WILL NOT MERGE ALL REFERENCE CONFS FOR SOME REASON
```
viktorklang, line 15: @dwijnand @jroper See this—I never seem to be able to get all the Serializer config I need from the ClassLoader, so I think this AutomaticFeature is dead in the water. As such, in this PR, you see that I added the serializers to the native-image reflecting config files manually.

proxy/core/src/main/scala/io/cloudstate/proxy/ProtobufGeneratedMessageRegisterFeature.scala
```diff
@@ -23,10 +23,20 @@ final class ProtobufGeneratedMessageRegisterFeature extends Feature {
     for {
```
viktorklang, line 23: The TCK passes using a native-image produced with this Protobuf AutomaticFeature—note that I had to filter specific package names to add methods for, as otherwise it would make the build take forever to run for some reason :S (most likely ballooning memory requirement).

build.sbt
```diff
@@ -375,14 +375,14 @@ def sharedNativeImageSettings(targetDir: File) = Seq(
     "akka.dispatch.affinity",
     "akka.util",
     "com.google.Protobuf",
+    "com.typesafe.config",
     "java.lang.ref.SoftReference", // https://github.com/oracle/graal/issues/2345
     "java.lang.invoke.MethodHandleImpl" // https://github.com/oracle/graal/issues/2345
   ).mkString("=", ",", ""),
-  "--initialize-at-run-time=" +
+  "-H:ClassInitialization=com.typesafe.config.impl.ConfigImpl$EnvVariablesHolder:rerun",
```
viktorklang, line 382: We can revert these to runtime once/if we decide to drop the serializer-automaticfeature. They are there to make sure that the system properties, environment variables, etc are loaded at runtime and not at build time.

PR #268, opened 2020-04-24
Introduce graal-tools for Graal-specific code

dwijnand:
Also makes it compile-time only and moves the code into its own package.


viktorklang:
Thanks @dwijnand!

PR #269, opened 2020-04-24
Hopefully fix overlap?

reid-spencer:
When viewed from cloudstate.io not github.com, a few of the items in the "Customizing" list at the end were overlapping on Google Chrome. This doesn't show up in github.com. Hopefully, putting the items on the same line solves this?


viktorklang:
Thanks @reid-spencer

pvlugter:
The format used created a description list (`dl`, `dt`, `dd` tags). Looks like the initial `datastore` term was missed in this change:

https://cloudstate.io/docs/user/deployment/deploying.html#customizing-the-stateful-service

Could be useful to fix the styling for these definition lists if that's broken.

PR #276, opened 2020-04-28
Reformat HOCON

dwijnand:
I wrote this a little while ago, as the default rendering in Typesafe Config makes a meal out of our list ConfigValue's.

The thinking was that if in the future we needed to run the guided, "generate me the JSON files" mode of native-image, then we could split that mega-file into pieces (I'll PR that utility code in a second) and format it sanely with this.

And/or this could be just wired up to the build to assist reformatting the config file and/or to CI to fail build validations.

Anyways.  Not wanting to lose this effort, I'm PR'ing it in case there's an interest.


dwijnand:
> then we could split that mega-file into pieces (I'll PR that utility code in a second)

https://github.com/cloudstateio/cloudstate/pull/277

PR #278, opened 2020-04-29
Docs kotlin updates

sleipnir:
Updated to the latest version 0.5.0


sleipnir:
ping @viktorklang @hseeberger @ChristopherAndre @marcellanz

sleipnir:
Someone here?

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
```diff
@@ -229,6 +242,58 @@ A minimal Maven example pom file, which uses the [Xolstice Maven Protocol Buffer
 ```
 @@@
 
+Gradle Kts
```
viktorklang, line 245: @sleipnir Perhaps put this in an external file and then include it in the docs via tags?

sleipnir: Thanks @viktorklang
I understand that it doesn't change much and I wouldn't have to worry about that. But okay, I can also change.

viktorklang: @sleipnir Makes it easier to notice and fix inadvertent changes and keep things up to date on chanes. Thanks!

sleipnir: Done!

viktorklang: @sleipnir Looks good—docs-test CI fails, you might want to have a look here: https://travis-ci.com/github/cloudstateio/cloudstate/jobs/327793454#L348

sleipnir: I've already seen that. The job running now is https://travis-ci.com/github/cloudstateio/cloudstate/builds/163811845

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
```diff
@@ -36,197 +36,35 @@ libraryDependencies += "io.cloudstate" % "cloudstate-kotlin-support" % "$cloudst
...
-                        <goals>
-                            <goal>compile</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
+Maven
+: @@@vars
+@@snip [Main.kt](/docs/src/test/kotlin/docs/user/gettingstarted/pom.xml) { #pom-xml }
+@@@
```
pvlugter, line 63: `@@@vars` is only needed when using paradox's variable substitution. Shouldn't be needed here.

```suggestion
: @@snip [Main.kt](/docs/src/test/kotlin/docs/user/gettingstarted/pom.xml) { #pom-xml }
```

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
```diff
@@ -36,197 +36,35 @@ libraryDependencies += "io.cloudstate" % "cloudstate-kotlin-support" % "$cloudst
...
-            </plugin>
-        </plugins>
-    </build>
-    
-</project>
-```
+Gradle Kts
+: @@@vars
+@@snip [Main.kt](/docs/src/test/kotlin/docs/user/gettingstarted/build.gradle.kts) { #build-kts }
 @@@
```
pvlugter, line 68: Same here, shouldn't need `@@@vars`

```suggestion
: @@snip [Main.kt](/docs/src/test/kotlin/docs/user/gettingstarted/build.gradle.kts) { #build-kts }
```

docs/src/test/kotlin/docs/user/gettingstarted/pom.xml
```diff
@@ -0,0 +1,181 @@
+// #pom-xml
+        <?xml version="1.0" encoding="UTF-8"?>
```
pvlugter, line 2: Just formatting, but expect this wasn't meant to be indented

```suggestion
<?xml version="1.0" encoding="UTF-8"?>
```

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
sleipnir: As the xml excerpt was removed, the library version remained in the file. How will the replacement occur in this case?

pvlugter: Okay, I didn't see any paradox variables in the pom.xml, and I don't think that paradox applies variable substitution to included snippets anyway.

What we've done before (in other projects using paradox) is run a variable substitution on build files first, which then get tested, and the generated files get pulled into the docs via snippets. We should add support for something like this. Will become obvious once we start testing the snippets anyway.

pvlugter: Captured this in an issue (#306) so we can follow up.

docs/src/test/kotlin/docs/user/gettingstarted/build.gradle.kts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+}
+
+dependencies {
+    implementation(kotlin("stdlib-jdk8"))
+    implementation("io.cloudstate:cloudstate-kotlin-support:$cloudstate.kotlin-support.version$")
```
pvlugter, line 18: Okay, see that the gradle build has a paradox variable. That won't be substituted.

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
pvlugter: Given that the other docs (like java and kotlin support) currently embed the build description samples so they can have var substitutions, I think it makes sense to just do that here for now too. So revert back to how this PR was originally. I can follow up with supporting a substitution step for files later.

sleipnir:
@pvlugter I returned the files

docs/src/main/paradox/user/lang/kotlin/gettingstarted.md
```diff
@@ -224,11 +237,65 @@ A minimal Maven example pom file, which uses the [Xolstice Maven Protocol Buffer
...
         </plugins>
     </build>
-    
+
 </project>
 ```
 @@@
 
+Gradle Kts
+<<<<<<< HEAD
```
pvlugter, line 246: Left over from merging

sleipnir: Fixed

sleipnir:
Thanks @pvlugter

PR #291, opened 2020-05-01
Diagnosed remaining assembly errors

raboof:
Only a couple remaining after https://github.com/akka/akka-persistence-spanner/pull/74
and https://github.com/akka/akka-grpc/pull/943


build.sbt
```diff
@@ -466,6 +471,21 @@ lazy val `proxy-spanner` = (project in file("proxy/spanner"))
...
+      // These are the only 2 .proto with this problem because the version in
+      // cloudstate has deviated from the upstream one since
+      // https://github.com/cloudstateio/cloudstate/pull/127
+      // We should still not ship these .proto's in akka-persistence-spanner, though
+      case "google/api/http.proto" => MergeStrategy.first
+      case "google/api/annotations.proto" => MergeStrategy.first
+      // This one is different in akka-protobuf-v3.
+      // We should still not ship these .proto's in akka-persistence-spanner
+      // and not in akka-protobuf-v3, though.
+      case "google/protobuf/field_mask.proto" => MergeStrategy.first
```
viktorklang, line 488: I think you can safely discard all ".proto"-files since the fat jar will not have them used for anything, it's only intended to be executed via java -jar

PR #293, opened 2020-05-02
Wip upgrade to 2.6.x √

viktorklang:
Lots of things to discuss here—but all-in-all this branch passes the TCK using the native-image generated.


build.sbt
```diff
@@ -49,30 +50,46 @@ name := "cloudstate"
...
 val DockerBaseImageJavaLibraryPath = "${JAVA_HOME}/lib"
 
-def excludeTheseDependencies = Seq(
+val excludeTheseDependencies: Seq[ExclusionRule] = Seq(
   ExclusionRule("io.netty", "netty"), // grpc-java is using grpc-netty-shaded
   ExclusionRule("io.aeron"), // we're using Artery-TCP
   ExclusionRule("org.agrona") // and we don't need this either
 )
 
+def akkaDependency(name: String, excludeThese: ExclusionRule*) =
```
viktorklang, line 70: I added these helpers to make maintenance easier—to make sure we exclude all we need at all times.

build.sbt
```diff
@@ -49,30 +50,46 @@ name := "cloudstate"
...
+  "com.typesafe.akka" %% name % AkkaPersistenceCassandraVersion excludeAll ((excludeTheseDependencies ++ excludeThese): _*)
+
 def common: Seq[Setting[_]] = Seq(
   headerMappings := headerMappings.value ++ Seq(
       de.heikoseeberger.sbtheader.FileType("proto") -> HeaderCommentStyle.cppStyleLineComment,
       de.heikoseeberger.sbtheader.FileType("js") -> HeaderCommentStyle.cStyleBlockComment
     ),
   // Akka gRPC overrides the default ScalaPB setting including the file base name, let's override it right back.
   akkaGrpcCodeGeneratorSettings := Seq(),
+  excludeDependencies += "com.typesafe.akka" %% "akka-protobuf-v3", // akka-protobuf-v3 shades protobuf-java 3.10.0 which is not compatible with native-image 20.0.0
```
viktorklang, line 92: We need to exclude this across the board—this was the easiest way I could think of.

build.sbt
```diff
@@ -306,6 +323,22 @@ def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
   }
 )
 
+def assemblySettings(jarName: String) =
```
viktorklang, line 326: There was lots of code duplication related to assembly that I factored out into this utility method.

build.sbt
```diff
@@ -306,6 +323,22 @@ def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
...
+  Seq(
+    mainClass in assembly := (mainClass in Compile).value,
+    assemblyJarName in assembly := jarName,
+    test in assembly := {},
+    // logLevel in assembly := Level.Debug,
+    assemblyMergeStrategy in assembly := {
+      /*ADD CUSTOMIZATIONS HERE*/
+      case PathList("META-INF", "io.netty.versions.properties") => MergeStrategy.last
+      case PathList(ps @ _*) if ps.last endsWith ".proto" => MergeStrategy.last
```
viktorklang, line 335: When we create "fat jars" it doesn't make sense to add the protos, as we don't expect anything to load them from the jar.

build.sbt
```diff
@@ -376,32 +413,30 @@ lazy val `proxy-core` = (project in file("proxy/core"))
         "io.grpc" % "grpc-netty-shaded" % GrpcJavaVersion,
         // Since we exclude Aeron, we also exclude its transitive Agrona dependency, so we need to manually add it HERE
         "org.agrona" % "agrona" % "0.9.29",
-        "com.typesafe.akka" %% "akka-remote" % AkkaVersion excludeAll (excludeTheseDependencies: _*),
+        akkaDependency("akka-remote"),
+        "com.typesafe.akka" %% "akka-protobuf-v390" % (AkkaVersion + "-proto390") excludeAll (excludeTheseDependencies: _*), // FIXME since akka-protobuf-v3 uses a shaded protobuf-java 3.10.0 which is not supported in native-image yet…
```
viktorklang, line 417: @patriknw special-cut this for us, as Protobuf-Java 3.10.0 uses VarHandles which are not fully working with GraalVM native-image 20.0.0 (and 20.1.0-dev).

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaSerializerRegisterFeature.scala
```diff
@@ -1,66 +0,0 @@
-package io.cloudstate.graaltools
```
viktorklang, line 1: This feature didn't work as intended, as it could not see what it needed to see, so I removed it until we can find a suitable replacement.

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -47,16 +54,17 @@ akka {
   cluster {
     shutdown-after-unsuccessful-join-seed-nodes = 60s
 
+    // FIXME consider settin distributed-data.max-delta-elements = <N>
```
viktorklang, line 57:
```suggestion
    // FIXME consider setting distributed-data.max-delta-elements = <N>
```

tck/src/it/resources/application.conf
```diff
@@ -1,3 +1,4 @@
+cloudstate-tck.verify = ["Akka + Node.js", "Akka + Java", "Akka + Scala", "Akka + Go", "Akka + Kotlin"]
```
viktorklang, line 1: I created this so that it is easier to specify which of the combinations one wants to run.

tck/src/it/scala/io/cloudstate/tck/TCK.scala
```diff
@@ -5,8 +5,22 @@ import com.typesafe.config.{Config, ConfigFactory}
...
-           toVector: _*) with SequentialNestedSuiteExecution
+class TCK extends Suites({
+  val config = ConfigFactory.load()
+  val combinations = config.getConfigList("cloudstate-tck.combinations")
+  val validNames = combinations.iterator.asScala.map(_.getString("name")).toSet
+  val verify = config.getStringList("cloudstate-tck.verify").asScala.toSet
+
+  verify.filterNot(validNames) match {
+    case x if !x.isEmpty =>
+      System.err.println("Configuration 'cloudstate-tck.verify' contains non-existent names for combinations:" + x.mkString("[",",","]"))
```
viktorklang, line 16:
```suggestion
      throw new IllegalArgumentException("Configuration 'cloudstate-tck.verify' contains non-existent names for combinations:" + x.mkString("[",",","]"))
```

build.sbt
```diff
@@ -16,6 +16,7 @@ inThisBuild(
     organizationHomepage := Some(url("https://lightbend.com")),
     startYear := Some(2019),
     licenses += ("Apache-2.0", new URL("https://www.apache.org/licenses/LICENSE-2.0.txt")),
+    resolvers += "Akka Snapshots" at "https://repo.akka.io/snapshots/", // FIXME for akka-protobuf-v390
```
hepin1989, line 19: So it needs a snapshot for quick prototyping?

build.sbt
```diff
@@ -49,30 +50,46 @@ name := "cloudstate"
...
-val AkkaVersion = "2.5.31"
+val AkkaVersion = "2.6.5"
 val AkkaHttpVersion = "10.1.11"
 val AkkaManagementVersion = "1.0.5"
 val AkkaPersistenceCassandraVersion = "0.102"
 val PrometheusClientVersion = "0.6.0"
 val ScalaTestVersion = "3.0.5"
-val ProtobufVersion = "3.9.0"
+val ProtobufVersion = "3.9.0" // We use this version because it is the latest which works with native-image 20.0.0
 val GraalVersion = "20.0.0"
```
hepin1989, line 60: 20.0.1?

viktorklang: @hepin1989 2.1.0 will be released in 2 weeks, so holding off making modifications to GraalVM until then.

proxy/core/src/graal/META-INF/native-image/com.typesafe/ssl-config/reflect-config.json.conf
```diff
@@ -0,0 +1,6 @@
+[
+  {
+    "name": "com.typesafe.sslconfig.ssl.NoopHostnameVerifier"
```
viktorklang, line 3: @jroper @pvlugter We need to decide on what to choose for our usage—I have no idea and the documentation on the matter was not very helpful either. Right now it loads the NoopHostnameVerifier.

build.sbt
viktorklang: @pvlugter @jroper @patriknw We need to decide on how to proceed with this—should we downgrade akka-protobuf to 3.9.0 or should we maintain an `akka-protobuf-v390` artifact, or should we consider having akka depending on (and not shading) protobuf-java 3?

pvlugter: It looks like the usage of method handles was removed again in protobuf-java 3.11. So we may be able to instead upgrade akka-protobuf-v3 to 3.11?

pvlugter: Also discussed in https://github.com/protocolbuffers/protobuf/issues/6497

build.sbt
pvlugter: 👍

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -47,16 +54,17 @@ akka {
...
 
     // Non-durable for now, since we can't get native-image to work with lmdb right now
     sharding.distributed-data.durable.keys = []
 
-    // fixme Of course, this is not ideal, but not much choice at the moment.
-    auto-down-unreachable-after = 30s
-
     sharding {
       rebalance-interval = 5s
+      passivate-idle-entity-after = off // FIXME put in a good value here
```
pvlugter, line 67: And remove the built-in passivation using receive timeout. Looks like `cloudstate.proxy.passivation-timeout` is currently 30s. Could substitute that setting here.

build.sbt
jroper: How does it work if they are excluded? Won't excluding this break all the serializers that depend on it? Note, since the TCK just uses a cluster of one, ie it doesn't do any remoting, any breakage caused by excluding these probably would not be picked up by the TCK.

jroper: Ah, I've now seen the comment below, disregard this.

proxy/core/src/graal/META-INF/native-image/com.typesafe/ssl-config/reflect-config.json.conf
jroper: Depends on what it's being used for.

graal-tools/src/main/scala/io/cloudstate/graaltools/ProtobufGeneratedMessageRegisterFeature.scala
```diff
@@ -11,9 +11,9 @@ import java.util.concurrent.ConcurrentHashMap
 final class ProtobufGeneratedMessageRegisterFeature extends Feature {
```
jroper, line 11: This should be modified. With the old version of protobuf that Akka used, the `equals`, `hashCode` and `toString` implementations used reflection, which is why every method needed a reflection configuration. But the new protobuf version they're on, when configured to be optimized for speed (rather than bytecode size), generates `equals`, `hashCode` and `toString` methods that don't use reflection. Furthermore, all of the `akka.protobufv3.internal.*` subtypes, Akka never loads them or interacts with them using reflection, rather, Akka has hand coded serializers that convert the Akka types to these types statically, and uses short manifest names, rather than classnames, with static mappings of these names to the protobuf types, when deserializisng. So, we should be able to just remove `akka.protobufv3.*` from here, no reflection configuration should be needed for it.

We still do need some config for the Cloudstate messages that we expect to be serialized by the Akka protobuf serializer, but I think there might be some problems with the code below, the Akka protobuf serializer uses `parseFrom` and `toByteArray` reflectively, but I don't see those being registered. Also, not sure how much of an overhead this is going to be to have configuration for classes that don't need it, we have a *lot* of protobuf classes that don't get loaded reflectively, only those sent over Akka remoting as top level messages need reflection configuration, which is only a handful of messages.

build.sbt
patriknw: I can try updating akka-protobuf-v3 to 3.11

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaSerializerRegisterFeature.scala
dwijnand: The issue here is that not all subtypes are reachable.  But for those that are, it's a convenience to register their constructor (which otherwise mightn't be reached by any bytecode).  (Side note: makes me wonder if registering the classes makes any sense, if the type is already reachable).

And for all the subtypes that are also used but not reached, I would just hard-code them here and reuse the same logic, instead of hard-code them verbosely in the config file.

build.sbt
```diff
@@ -334,17 +369,19 @@ def sharedNativeImageSettings(targetDir: File) = Seq(
     "akka.dispatch.affinity",
     "akka.util",
     "com.google.Protobuf",
-    "com.typesafe.config",
-    "java.lang.ref.SoftReference", // https://github.com/oracle/graal/issues/2345
-    "java.lang.invoke.MethodHandleImpl" // https://github.com/oracle/graal/issues/2345
+    "scala.runtime.Statics$VM"
```
dwijnand, line 372: I would've thought this was redundant with the `"scala"` above.

build.sbt
```diff
@@ -851,13 +835,14 @@ lazy val `graal-tools` = (project in file("graal-tools"))
   .enablePlugins(GraalVMPlugin)
   .settings(
     libraryDependencies ++= List(
-        "org.graalvm.nativeimage" % "svm" % GraalVersion,
+        "org.graalvm.sdk" % "graal-sdk" % GraalVersion % "provided",
```
dwijnand, line 838: Huh.  What happened here to bring this back?

viktorklang: Ah, good question, I had to add that when trying the 20.1.0-dev build since it wasn't officially published anywhere. Can remove this now.

build.sbt
viktorklang: @dwijnand I'll see if that's the case

graal-tools/src/main/scala/io/cloudstate/graaltools/ProtobufGeneratedMessageRegisterFeature.scala
viktorklang: @jroper I'll try removing protobufv3 and report back

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang: @pvlugter I suspect that we need to remove the receive timeout thing at the same time to not have competing timers?

build.sbt
patriknw: https://github.com/akka/akka/pull/29014

graal-tools/src/main/scala/io/cloudstate/graaltools/ProtobufGeneratedMessageRegisterFeature.scala
viktorklang: @jroper You were right! Removing the protobufv3 worked. I'll see if I can remove the need for the methods again.

graal-tools/src/main/java/io/cloudstate/graaltools/Substitutions.java
```diff
@@ -42,43 +42,6 @@
...
 
-@TargetClass(
-    className =
-        "io.grpc.netty.shaded.io.netty.util.internal.shaded.org.jctools.util.UnsafeRefArrayAccess")
-final
-class Target_io_grpc_netty_shaded_io_netty_util_internal_shaded_org_jctools_util_UnsafeRefArrayAccess {
-  @Alias
-  @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.ArrayIndexShift, declClass = Object[].class)
-  public static int REF_ELEMENT_SHIFT;
-}
```
dwijnand, line 53: Btw, these were from `proxy/cassandra/src/main/java/io/cloudstate/internal/svm/Substitutions.java`.  Does that still build without these?

viktorklang: @dwijnand GraalVM 20.0.0 adds these Substitutions itself.

viktorklang:
@pvlugter This is ready to merge on my end.

graal-tools/src/main/scala/io/cloudstate/graaltools/ProtobufGeneratedMessageRegisterFeature.scala
viktorklang: @jroper It would be interesting to simplyfy this AutomaticFeature—but the challenge is the ridiculously long turnaround time to try even the smallest changes. Feel free (encouraged!) to find optimizations here! :+1:

pvlugter:
> This is ready to merge on my end

My preference would be to keep this open until the WIP and FIXMEs are addressed, so they're not forgotten or become comfortable hanging out in the codebase.

proxy/core/src/main/resources/cloudstate-common.conf
pvlugter: Yes, remove the receive timeout at the same time. Although, I can imagine that passivation timeouts is something that will want to be defined per entity type on the user side? Maybe this needs its own issue to be enhanced.

tck/src/it/resources/application.conf
pvlugter: Needs to be added to `native-image.conf` as well, otherwise the native TCK integration test won't run.

build.sbt
viktorklang, line 19: @pvlugter This cannot be fixed until Akka 2.6.6

build.sbt
viktorklang: @pvlugter This cannot be fixed until Akka 2.6.6

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -3,6 +3,9 @@
 akka {
   actor {
     provider = cluster
+    allow-java-serialization = off
+    warn-about-java-serializer-usage = off
+    internal-dispatcher = akka.actor.default-dispatcher // FIXME consider removing this line
```
viktorklang, line 8: @pvlugter @patriknw Probably best solved by a decision—given that we control all that runs in this, I thouht it important to keep number of threads down. If you both agree with this setting, then I can remove the comment.

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -39,6 +42,8 @@ akka {
     artery {
       enabled = on
       transport = tcp
+      // FIXME do we need to set canonical.hostname?
+      // FIXME do we need to set advanced.tcp.connection-timeout?
```
viktorklang, line 46: @pvlugter @patriknw Should we set these two new settings, and to what?

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -47,16 +52,17 @@ akka {
   cluster {
     shutdown-after-unsuccessful-join-seed-nodes = 60s
 
+    // FIXME consider setting distributed-data.max-delta-elements = <N>
+    // FIXME consider setting distributed-data.delta-crdt.max-delta-size = <N>
```
viktorklang, line 56: @pvlugter @patriknw These are here since they are a part of the migration docs, and I didn't want them to be forgotten, but I also don't know what to set them to. Advise?

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang: @pvlugter Makes sense to make the change separately from this PR. (Hence the FIXME)

tck/src/it/resources/application.conf
viktorklang: @pvlugter Fixed.

viktorklang:
@pvlugter There are no wip/fixmes in here that can currently be addressed (barring decisions needed)

proxy/core/src/main/resources/cloudstate-common.conf
patriknw: I think it's a valid choice for Cloudstate

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -39,6 +42,8 @@ akka {
     artery {
       enabled = on
       transport = tcp
```
patriknw, line 44: both these are default in Akka 2.6, so not strictly needed

proxy/core/src/main/resources/cloudstate-common.conf
```diff
@@ -3,6 +3,9 @@
 akka {
   actor {
     provider = cluster
+    allow-java-serialization = off
```
patriknw, line 6: default in Akka 2.6, so not strictly needed

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang: Thanks @patriknw!

viktorklang, line 8:
```suggestion
    internal-dispatcher = akka.actor.default-dispatcher
```

proxy/core/src/graal/META-INF/native-image/com.typesafe/ssl-config/reflect-config.json.conf
viktorklang: @patriknw Do you have any suggestion what what hostname verifier we should be using?

proxy/core/src/main/resources/cloudstate-common.conf
patriknw: default connection-timeout is 5s, so unless you think that is wrong you should start with defaults

patriknw: if you don't set hostname it will be `InetAddress.getLocalHost.getHostAddress`, maybe good to have that as an optional environment variable here?

proxy/core/src/main/resources/cloudstate-common.conf
patriknw: Remove, and use defaults until you see a problem.
Only mentioned in migration guide because the defaults were changed (to better defaults).

proxy/core/src/graal/META-INF/native-image/com.typesafe/ssl-config/reflect-config.json.conf
patriknw: This is sslconfig. That isn't used for Akka remoting. It was deprecated in Akka 2.6. I would be surprised if it's used by Akka gRPC. @raboof do you know? Or is it still used in Akka HTTP?

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang, line 6:
```suggestion
```

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang, line 44:
```suggestion
```

proxy/core/src/main/resources/cloudstate-common.conf
viktorklang, line 56:
```suggestion
```

proxy/core/src/main/resources/cloudstate-common.conf
pvlugter: Sure. Would be good to capture in an issue then, rather than just a FIXME.

pvlugter:
Okay, captured a couple more follow-up issues in #297 and #298.

PR #299, opened 2020-05-06
Always use --no-server for native-image in docker

pvlugter:
Trying the native image building in docker, it didn't have enough memory allocated by default. I'm assuming that always disabling the build server is good for the docker build anyway, and it then defaults to using 80% of the available memory (and also allows heap size to be set if needed).

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -310,7 +310,7 @@ def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
         case Some(_) => new File("/opt/docker/graal-resources/")
         case None => baseDirectory.value / "src" / "graal"
       }
-    }),
+    }, graalVMBuildServer.value),
   dockerEntrypoint := {
     val old = dockerEntrypoint.value
     val withLibraryPath = if (graalVMDockerPublishLocalBuild.value) {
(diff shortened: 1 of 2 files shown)


build.sbt
```diff
@@ -305,11 +305,13 @@ def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
...
-  graalVMNativeImageOptions ++= sharedNativeImageSettings({
-      graalVMVersion.value match {
-        case Some(_) => new File("/opt/docker/graal-resources/")
-        case None => baseDirectory.value / "src" / "graal"
-      }
+  graalVMNativeImageOptions ++= (graalVMVersion.value match {
+      case Some(_) =>
+        sharedNativeImageSettings(new File("/opt/docker/graal-resources/")) ++ Seq(
+          // disable build server for docker by default (and use 80% of available memory)
+          "--no-server"
```
viktorklang, line 312: May I suggest we try to bake this into the sharedNativeImageSettings method—otherwise I think people might get confused if settings are added in multiple places.

pvlugter: Okay, sure

pvlugter: Updated to be a switch on the shared settings, and added an sbt setting for this so it can be set for local builds without needing to (un)comment.

PR #302, opened 2020-05-06
Make the build use a Java11 build image for native-image!

### LINKED ISSUE #301: Use Java 11 for GraalVM native-image
Currently we are using the tag `20.0.0` for the native-image Docker image in our `build.sbt`. This refers to Java 8!

In order to use Java 11 we have to change to `20.0.0-java11`:

``` scala
val GraalVersion = "20.0.0-java11"
```

viktorklang:
Resolves #301

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -304,7 +304,7 @@ commands ++= Seq(
 def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
   // If this is Some(…): run the native-image generation inside a Docker image
   // If this is None: run the native-image generation using a local GraalVM installation
-  graalVMVersion := Some(GraalVersion),
+  graalVMVersion := Some(GraalVersion + "-java11"), // make sure we use the java11 version
   graalVMNativeImageOptions ++= sharedNativeImageSettings({
       graalVMVersion.value match {
         case Some(_) => new File("/opt/docker/graal-resources/")


hseeberger:
See #301.

build.sbt
```diff
@@ -305,6 +305,8 @@ def nativeImageDockerSettings: Seq[Setting[_]] = dockerSettings ++ Seq(
   // If this is Some(…): run the native-image generation inside a Docker image
   // If this is None: run the native-image generation using a local GraalVM installation
   graalVMVersion := Some(GraalVersion),
+  graalVMContainerBuildImage := graalVMContainerBuildImage.value
+      .map(_ + "-java11"), // We want to use the Java 11 version of GraalVM
```
hseeberger, line 309: I think this will work, but I don not understand the difference btw. this approach and the one I suggested (set `GraalVersion` to `something-java11`.

viktorklang: @hseeberger The Defense presents exhibit A: https://github.com/cloudstateio/cloudstate/blob/master/build.sbt#L836

hseeberger: Accepted ;-)

pvlugter: Doesn't seem to work — I still see it build with java8 version. The container build image is already created at this point, it just modifies its name. Could just do this instead?

```
graalVMVersion := Some(GraalVersion + "-java11")
```

viktorklang: @pvlugter Done!

pvlugter:
👍

PR #310, opened 2020-05-07
It would appear that the first Chunk is always discarded, this fixes …

viktorklang:
…that

### DIFF
--- a/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
+++ b/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
@@ -517,36 +517,33 @@ object HttpApi {
         )
 
       if (methDesc.isServerStreaming) {
+        def toChunk(entityMessage: MessageOrBuilder): HttpEntity.Chunk =
+          HttpEntity.Chunk(
+            if (isHttpBodyResponse) extractDataFromHttpBody(entityMessage) ++ NEWLINE_BYTES
+            else ByteString(jsonPrinter.print(entityMessage)) ++ NEWLINE_BYTES
+          )
+
         data
           .prefixAndTail(1)
           .runWith(Sink.head)
           .map {
             case (Seq(protobuf: ProtobufAny), rest) =>
-              val entityMessage: MessageOrBuilder = parseResponseBody(Serve.ReplySerializer.serialize(protobuf))
+              val entityMessage: MessageOrBuilder = parseResponseBody(protobuf)
               val contentType =
                 if (isHttpBodyResponse) extractContentTypeFromHttpBody(entityMessage)
                 else ContentTypes.`application/json`
 
-              val first = {
-                val data =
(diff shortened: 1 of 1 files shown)


sleipnir:
It seems worthwhile to open an issue to the AKKA repository for research right?

viktorklang:
@sleipnir Not really, this is our code (Cloudstate).

sleipnir:
Okay @viktorklang I proposed this because I already had problems using FileIO where it dropped some messages, but I have no evidence of that at the moment, so I thought it could be something more profound.

viktorklang:
@sleipnir Check if it fares better with this PR :)

ralphlaude:
@viktorklang, cool. How did you find it out?

viktorklang:
@ralphlaude Pure luck, while reading code.

pvlugter:
So this is not tested in any way?

viktorklang:
@pvlugter Let me check, perhaps it is not checking the right thing.

raboof:
I'll port this one since it conflicts with #312

proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
```diff
@@ -527,22 +527,26 @@ object HttpApi {
...
 
-              val data =
-                if (isHttpBodyResponse) extractDataFromHttpBody(entityMessage)
-                else ByteString(jsonPrinter.print(entityMessage))
+              val first = {
+                val data =
+                  if (isHttpBodyResponse) extractDataFromHttpBody(entityMessage)
+                  else ByteString(jsonPrinter.print(entityMessage))
+
+                Source.single(HttpEntity.Chunk(data))
```
raboof, line 535: no newline here

proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
```diff
@@ -527,22 +527,26 @@ object HttpApi {
...
               HttpResponse(
                 entity = HttpEntity.Chunked(
                   contentType,
-                  rest map { protobuf: ProtobufAny =>
+                  first.concat(rest map { protobuf: ProtobufAny =>
                     val entityMessage: MessageOrBuilder = parseResponseBody(Serve.ReplySerializer.serialize(protobuf))
                     val data =
                       if (isHttpBodyResponse) extractDataFromHttpBody(entityMessage)
                       else ByteString(jsonPrinter.print(entityMessage))
                     HttpEntity.Chunk(data ++ NEWLINE_BYTES)
```
raboof, line 548: but a newline here. is that OK?

proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
viktorklang: Good catch—addressing

PR #311, opened 2020-05-08
Refactor: simplify HttpApi transformResponse

raboof:
By pulling out the 'unwrapping' logic to a helper function.
Makes it easier to digest I think. The same could probably be done
for the streaming response case, but I didn't want to create a conflict
with #310 ;)


raboof:
(thanks, superseded by #312 though :) )

PR #312, opened 2020-05-08
Refactor proxy: don't convert to HttpResponse prematurely

raboof:
By keeping the data as `Source[ProtobufAny]` instead of converting it to a 
HttpResponse immediately after receiving it, we can save some
back-and-forth interpreting the data.

### DIFF
--- a/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
+++ b/proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
@@ -25,6 +25,7 @@ import scala.util.matching.Regex
 import scala.util.parsing.combinator.Parsers
 import scala.util.parsing.input.{CharSequenceReader, Positional}
 import akka.ConfigurationException
+import akka.NotUsed
 import akka.http.scaladsl.model.{
   ContentType,
   ContentTypes,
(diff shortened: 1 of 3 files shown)


proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
```diff
@@ -491,110 +493,76 @@ object HttpApi {
...
+        entityMessage
+          .getField(entityMessage.getDescriptorForType.findFieldByName("content_type")) match {
+          case null | "" => ContentTypes.NoContentType
+          case string: String =>
+            ContentType.parse(string) match {
+              case Left(list) =>
+                throw new IllegalResponseException(
+                  list.headOption.getOrElse(ErrorInfo.fromCompoundString("Unknown error"))
+                )
+              case Right(tpe) => tpe
```
viktorklang, line 507:
```suggestion
            ContentType.parse(string).fold(list =>
                throw new IllegalResponseException(
                  list.headOption.getOrElse(ErrorInfo.fromCompoundString("Unknown error"))
                ), identity)
```

viktorklang:
Thanks @raboof!

PR #315, opened 2020-05-12
Auto-configure Akka's serializers for native-image

### LINKED ISSUE #271: Automate Serializer reflection configuration for native-image
Right now we have static reflection configuration for the Serializers needed for our native-image. We have a prototype, which unfortunately does not work, because the needed *.conf files we need to merge and read from are not visible during analysis—hence it is currently not enabled.

So the task is to get this AutomaticFeature working: https://github.com/cloudstateio/cloudstate/blob/master/graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaSerializerRegisterFeature.scala

dwijnand:
Fixes #271


project/GraalVMPlugin.scala
```diff
@@ -226,6 +228,24 @@ object GraalVMPlugin extends AutoPlugin {
     toMapping(files).map(_._2)
   }
 
+  private val mkDynamicAkkaReflectConfig = Def.task {
```
viktorklang, line 231: Wow—nice and clean code, @dwijnand! :+1:

viktorklang:
Thanks @dwijnand!

PR #319, opened 2020-05-13
Update Akka gRPC to latest snapshot, alpn-agent to .10

raboof:
Make it work with JDK 8u252. Not tested beyond 'sbt test' yet.


hseeberger:
Thanks @raboof. I think we are on Java 11 now, so is this needed? /cc @pvlugter

raboof:
> is this needed

For JDK 11 it's not so urgent.

Still, the behavior changes on JDK 11 as well (the netty gRPC client is allowed to select the native OpenSSL implementation instead of using the JDK one), so I'd like to make sure that works for cloudstate before releasing gRPC 1.0.0 ;).

viktorklang:
@raboof Let me know when this is ready to merge

raboof:
Marking 'ready to review', good to go from my side, tested with native-image locally. Would be great not to have to downgrade grpc-netty-shaded, but that can be a separate task.

build.sbt
```diff
@@ -48,10 +48,15 @@ def versionFmt(out: sbtdynver.GitDescribeOutput): String = {
 
 name := "cloudstate"
 
-val GrpcJavaVersion = "1.22.1"
+// Should be in sync with the version used in Akka gRPC
+val GrpcJavaVersion = "1.29.0"
+// Unfortunately we need to downgrade grpc-netty-shaded
+// in the proxy until we have a fix to make it work with
+// native-image
+val GrpcNettyShadedVersion = "1.28.1"
```
pvlugter, line 56: This doesn't seem to be used anywhere, so grpc-netty-shaded is not downgraded?

raboof: Oops - I'm fairly sure I just forgot to commit that and it's actually needed for the native-image tck test - but I'll double-check.

PR #320, opened 2020-05-13
Update Akka HTTP to 10.1.12

raboof:
Not strictly important right now, but makes the delta of upgrading
Akka gRPC smaller later ;)


hseeberger:
Thanks. LGTM.

PR #321, opened 2020-05-13
Rework ProtobufGeneratedMessageRegisterFeature

### LINKED ISSUE #295: Reviste configuration generation for Protobufs + native-image
Right now we're using [ProtobufGeneratedMessageRegisterFeature](https://github.com/cloudstateio/cloudstate/blob/master/graal-tools/src/main/scala/io/cloudstate/graaltools/ProtobufGeneratedMessageRegisterFeature.scala)—but I suspect that it is not going to be manageable: we're going to over-configure or risk not configuring all that needs to be configured. So we need to come up with some other solution to the problem.

dwijnand:
Rather than registering subclasses of various Protobuf message classes,
registering the nested classes within key Protobuf classes seems
sufficient.  Also, this can all happen once, before analyis, rather than
during analysis, which requires the cache.

Also, rework AkkaActorRegisterFeature to setup a
registerSubtypeReachabilityHandler in "beforeAnalysis", which also drops
the need for a cache.

Finally, drop a few more manual config points, which I was experimenting
with initialising on build (an abandoned approach) and seemingly turned
out not to be necessary anyways.

Fixes #295

### DIFF
--- a/graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
+++ b/graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
@@ -1,30 +1,25 @@
 package io.cloudstate.graaltools
 
 import com.oracle.svm.core.annotate.AutomaticFeature
-import org.graalvm.nativeimage.hosted.Feature.QueryReachabilityAccess
 import org.graalvm.nativeimage.hosted.{Feature, RuntimeReflection}
-
-import scala.collection.JavaConverters._
-import java.util.concurrent.ConcurrentHashMap
 
 @AutomaticFeature
 final class AkkaActorRegisterFeature extends Feature {
-  private[this] final val cache = {
-    val c = ConcurrentHashMap.newKeySet[String]
+  private[this] final val ignoreClass = Set(
     // FIXME the following are configured in akka-graal-config, remove this when migrating from it
-    c.add("akka.io.TcpConnection")
-    c.add("akka.io.TcpListener")
-    c.add("akka.stream.impl.fusing.ActorGraphInterpreter")
-    c
-  }
+    "akka.io.TcpConnection",
+    "akka.io.TcpListener",
+    "akka.stream.impl.fusing.ActorGraphInterpreter"
+  ).contains(_)
 
-  override final def duringAnalysis(access: Feature.DuringAnalysisAccess): Unit = {
+  override final def beforeAnalysis(access: Feature.BeforeAnalysisAccess): Unit = {
     val akkaActorClass =
       access.findClassByName(classOf[akka.actor.Actor].getName) // We do this to get compile-time safety of the classes, and allow graalvm to resolve their names
-    if (akkaActorClass != null && access.isReachable(akkaActorClass)) {
+
+    def register(subtype: Class[_]) =
       for {
-        subtype <- access.reachableSubtypes(akkaActorClass).iterator.asScala
-        if subtype != null && !subtype.isInterface && cache.add(subtype.getName)
+        subtype <- Option(subtype)
+        if !subtype.isInterface && ignoreClass(subtype.getName)
         _ = println("Automatically registering actor class for reflection purposes: " + subtype.getName)
         _ = RuntimeReflection.register(subtype)
         _ = RuntimeReflection.register(subtype.getDeclaredConstructors: _*)
@@ -34,7 +29,8 @@ final class AkkaActorRegisterFeature extends Feature {
       } {
         RuntimeReflection.register( /* finalIsWritable = */ true, context, self)
       }
-    }
+
+    access.registerSubtypeReachabilityHandler((_, subtype) => register(subtype), akkaActorClass)
   }
 
   private[this] final def getDeclaredField(cls: Class[_], name: String) =
--- a/graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
+++ b/graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
@@ -19,7 +19,7 @@ final class AkkaActorRegisterFeature extends Feature {
     def register(subtype: Class[_]) =
       for {
         subtype <- Option(subtype)
-        if !subtype.isInterface && ignoreClass(subtype.getName)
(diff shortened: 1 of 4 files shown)


viktorklang:
@dwijnand Cool—how does this affect native-image memory usage and/or build time?

dwijnand:
It's about the same.

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
```diff
@@ -1,30 +1,25 @@
...
     val akkaActorClass =
       access.findClassByName(classOf[akka.actor.Actor].getName) // We do this to get compile-time safety of the classes, and allow graalvm to resolve their names
-    if (akkaActorClass != null && access.isReachable(akkaActorClass)) {
+
+    def register(subtype: Class[_]) =
       for {
-        subtype <- access.reachableSubtypes(akkaActorClass).iterator.asScala
-        if subtype != null && !subtype.isInterface && cache.add(subtype.getName)
+        subtype <- Option(subtype)
+        if !subtype.isInterface && ignoreClass(subtype.getName)
```
pvlugter, line 22: Looks like this should be `!ignoreClass`?

```suggestion
        if !subtype.isInterface && !ignoreClass(subtype.getName)
```

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
```diff
@@ -1,30 +1,25 @@
...
     // FIXME the following are configured in akka-graal-config, remove this when migrating from it
-    c.add("akka.io.TcpConnection")
-    c.add("akka.io.TcpListener")
-    c.add("akka.stream.impl.fusing.ActorGraphInterpreter")
-    c
-  }
+    "akka.io.TcpConnection",
+    "akka.io.TcpListener",
+    "akka.stream.impl.fusing.ActorGraphInterpreter"
+  ).contains(_)
```
pvlugter, line 13: Set.apply calls contains, so not needed?

```suggestion
  )
```

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
dwijnand: Uuuhhh.... how does it work with that mistake in place? 🤯

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
dwijnand: Yeah, it does, I prefer to explicitly make it a predicate (it's this whole "intentional vs extension" set thing, which I won't go into).  Oh, and this way was terser than type annotating it as a function.  Anyways, happy to drop it.

graal-tools/src/main/scala/io/cloudstate/graaltools/AkkaActorRegisterFeature.scala
viktorklang: @dwijnand How are you testing that it works?

dwijnand: `proxy-core/graalvm-native-image:packageBin; tck/it:test` together with

```diff
--- tck/src/it/resources/application.conf
+++ tck/src/it/resources/application.conf
@@ -1,4 +1,4 @@
-cloudstate-tck.verify = ["Akka + Node.js", "Akka + Java", "Akka + Scala", "Akka + Go", "Akka + Kotlin"]
+cloudstate-tck.verify = ["Akka + Java"]
 cloudstate-tck.combinations = [{
   name = "Akka + Node.js"
   tck {
@@ -44,8 +44,8 @@ cloudstate-tck.combinations = [{
     hostname = "127.0.0.1"
     port = 9000
     directory = ${user.dir}
-    command = ["java","-Xmx512M", "-Xms128M", "-Dconfig.resource=in-memory.conf", "-Dcloudstate.proxy.dev-mode-enabled=true", "-jar", "proxy/core/target/scala-2.12/akka-proxy.jar"]
-    //command = ["proxy/core/target/graalvm-native-image/cloudstate-proxy-core", "-Djava.library.path="${JAVA_HOME}"/lib", "-Dconfig.resource=in-memory.conf", "-Dcloudstate.proxy.dev-mode-enabled=true"]
+    //command = ["java","-Xmx512M", "-Xms128M", "-Dconfig.resource=in-memory.conf", "-Dcloudstate.proxy.dev-mode-enabled=true", "-jar", "proxy/core/target/scala-2.12/akka-proxy.jar"]
+    command = ["proxy/core/target/graalvm-native-image/cloudstate-proxy-core", "-Djava.library.path="${JAVA_HOME}"/lib", "-Dconfig.resource=in-memory.conf", "-Dcloudstate.proxy.dev-mode-enabled=true"]

     env-vars {
       USER_FUNCTION_PORT = "8090"
```

viktorklang: That should work... What happens if you omit "-Dcloudstate.proxy.dev-mode-enabled=true"?

pvlugter: It fails when running with the docker build:

```
> dockerBuildNativeDevMode publishLocal
...
Error: A field that was already processed by the analysis cannot be re-registered as writable: private final akka.actor.ActorContext akka.io.TcpConnection.context
```

pvlugter: Same if I use a local build:

```
❯ set `proxy-core`/graalVMVersion := None
❯ proxy-core/graalvm-native-image:packageBin
...
[error] Error: A field that was already processed by the analysis cannot be re-registered as writable: private final akka.actor.ActorContext akka.io.TcpListener.context
```

pvlugter:
Should also work with the graal upgrade on master. Native image integration test running in CircleCI is enabled on branches now, so that will test it too.

pvlugter:
Green build in CircleCI: https://circleci.com/gh/cloudstateio/cloudstate/340

PR #326, opened 2020-05-19
Add AOT-compiled versions of proxy docker builds

pvlugter:
We've been experimenting with AOT-compilation with jaotc, as a possible alternative to native-image. We'll have performance details from load testing later on. Drafting the build changes for now to make them available.


pvlugter:
We'll likely stick with native images for a while. Closing this, but we can always revisit with the changes here to explore ahead-of-time compiling + class data sharing for faster startup of a hotspot JVM.

PR #329, opened 2020-05-20
Fixed an infinite recursion in Java support library

jroper:


viktorklang:
@pvlugter Indeed. Let's organize some test-infra spikes so that fleshing out the test support has a low-as-possible barrier to entry.

PR #336, opened 2020-05-24
Fix tabbed code blocks with empty lines in kotlin docs

pvlugter:


pvlugter:
I've checked this locally. Will merge so the docs can build.

PR #337, opened 2020-05-25
Use a slimmer docker base image

pvlugter:
Switch to the debian slim with just JRE docker image. Not sure if there was a reason for using a full JDK image, but shouldn't be needed at runtime. This takes around 250MB off the images. There are probably possibilities to go smaller again, especially for native images, but this is an easy first step.


viktorklang:
@pvlugter Let's see if this works!

sleipnir:
It would be interesting in native images that they were compiled statically, that way we could generate even smaller images with busybox, alpine, or other smaller base image

Issue #104, opened 2019-09-17
Polyglot docs


jroper:
Eventually, all or most of the language specific support libraries may live in other repos, with Go likely being the first to do this, as discussed in #103. This raises a question for the docs. In an ideal world, documentation should live in the same repository as the code it documents, since updating documentation when a feature is added or changed should be part of the definition of done of a task, and PRs shouldn't be merged until the feature is done. It also allows things like code snippet validation to capture breaking changes before they are committed, and guarantee that if the breaking change is acceptable, the docs get updated since CI doesn't pass without it. But right now our approach to the docs is in a mono structure, with a single paradox project, and this also has a lot of advantages.

I think splitting up the docs will probably be the best answer long term, this will mean coming up with a way to deploy the docs that allows them to be split across repos. It might mean instead of deploying the docs to a single repo (replacing the entire contents of the repo each time), we deploy them to a subdirectory within a shared repo. We currently use travis's GitHub pages deployment support to deploy docs, that probably won't work, but we can easily replace this with our own custom scripts.

ihostage:
@jroper Antora how one of the variants. I seeing that Tim worked on migrate Lightbend Tech Hub to Antora. 😄

marcellanz:
@jroper Would that apply to shared protobuf files for the CloudState protocol too? Together with the documentation they are shared with the main CloudState repository. If a support library builds in its repository by itself, it would have to download/copy over from somewhere. Would that be a similar shared artefact?

pvlugter:
I've created a shared paradox theme and plugin at https://github.com/cloudstateio/cloudstate-paradox.

Updated the main documentation to use this in #339.

For publishing the separate docs for language support modules: since we're using github pages, a simple approach would be publishing docs to the `gh-pages` branch of each support repo. These are automatically served under the repo name, so for example, the go support docs would be served under `https://cloudstate.io/go-support`. That may not be the structure that we want, as this is not under `/docs/`.

We could update the structure of the docs and publish everything under the one https://github.com/cloudstateio/docs repo. So we could change to having something like: `cloudstate.io/docs/core`, `cloudstate.io/docs/go`, and so on. This would be a breaking change for any links out there.

[...]

PR #339, opened 2020-05-26
Switch to cloudstate-paradox plugin and theme

pvlugter:
As part of #104, have moved the Cloudstate theme to a separate project, with an sbt plugin, so that this can be shared across the main repo and language support modules:

https://github.com/cloudstateio/cloudstate-paradox

Switch over to the plugin and published theme, which produces the same docs as before.


marcellanz:
nice @pvlugter. I like the concept of having versioned docs as well as having each support library its own documentation to be produced, especially as it helps to have tested code samples.

I find it compelling to have the ability to switch languages while reading the docs. `core` or `common` documentation would be the same for every language, but if one wants to read how to register a Cloudstate entity or how to get a CRDT entity initialised, switching languages without loosing context by opening a new location would be great I think. What is the thinking or experience here from the project and others?

akka.io similarly has the perhaps the same approach to switch between Java and Scala? I don't know akka.io docs well yet, but with your last note, would this enable to have a switch in the page for different languages (see screen below) or is this the same paradox source at akka.io? (inspecting the page source seems to indicate that).

Depending on how to switch languages, so not only switch between a code snipped but documentation around it, might make this a bit harder to achieve?

---
language switch in akka.io:

Screenshot 2020-05-26 at 12 04 33

pvlugter:
Merging this, to test the docs deployment properly, and so the docs can be published under the new locations.

PR #348, opened 2020-06-04
Upgrading to the buildin Base64 encode/decode in ByteString

viktorklang:


proxy/core/src/main/scala/io/cloudstate/proxy/GrpcWebSupport.scala
```diff
@@ -120,15 +120,12 @@ object GrpcWebSupport {
...
-        HttpEntity.Chunked.fromData(contentType, streamed.dataBytes.map(base64Encode))
+        HttpEntity.Chunked.fromData(contentType, streamed.dataBytes.map(_.encodeBase64))
     }
   }
 
-  private def base64Encode(bytes: ByteString): ByteString =
-    ByteString(Base64.getEncoder.encode(bytes.asByteBuffer))
-
   private def base64Decode(bytes: ByteString): ByteString =
     ByteString(Base64.getDecoder.decode(bytes.asByteBuffer))
```
pvlugter, line 130: Remove the decode method as well?

viktorklang: Good catch, thanks @pvlugter!

PR #349, opened 2020-06-05
Add basic instrumentation for event-sourced entities

pvlugter:
Add some simple metrics for event-sourced entities. Just counts currently, no timing metrics. Uses the prometheus client, since that's already available and exported. Counts can be converted to rates in prometheus queries.

Metrics are only tagged by the user function entity name (such as `"shopping-cart"`) so they're aggregated counts for an entity type on a node. Per-instance metrics are expected to be too high cardinality. Metrics include:

* gauge: entity instances currently active
* count: entity instances that have activated
* count: entity instances that have passivated
* count: commands received for an entity
* count: commands stashed for an entity
* count: commands processed for an entity
* count: events persisted for an entity
* count: event sizes (bytes) persisted for an entity
* count: events loaded for an entity
* count: event sizes (bytes) loaded for an entity
* count: snapshots persisted for an entity
* count: snapshot sizes (bytes) persisted for an entity
* count: snapshots loaded for an entity
* count: snapshot sizes (bytes) loaded for an entity

Only tested manually at the moment, can look at automated testing for the instrumentation later.

### DIFF
--- a/proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
+++ b/proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
@@ -32,6 +32,7 @@ import io.cloudstate.protocol.event_sourced._
 import io.cloudstate.proxy.ConcurrencyEnforcer.{Action, ActionCompleted}
 import io.cloudstate.proxy.StatsCollector
 import io.cloudstate.proxy.entity.{EntityCommand, UserFunctionReply}
+import io.cloudstate.proxy.telemetry.CloudstateTelemetry
 
 import scala.collection.immutable.Queue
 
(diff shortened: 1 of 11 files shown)


proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
```diff
@@ -179,6 +193,7 @@ final class EventSourcedEntity(configuration: EventSourcedEntity.Configuration,
 
   private[this] final def commandHandled(): Unit = {
     currentCommand = null
+    instrumentation.commandProcessed()
```
hseeberger, line 196: Not sure if it matters, but shouldn't "processed" (past tense) be called at the end of the code block which does the processing? Similar applies to `commandStashed` etc. below.

viktorklang:
Great stuff, @pvlugter!

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
pvlugter: This is currently after the command has been fully processed asynchronously, including persisting events. It's in the `commandHandled` method, after the current command has been cleared and before any stashed commands are processed. So you're suggesting having a counter after it sends the command to the user function (at the end of `handleCommand`)? Or should we change the verb for this hook if `process` doesn't work? We often used `completed` for this kind of event in Cinnamon.

I'll update the stashed hook to be a line down, after the stash, and also add an unstashed hook — so we can see the number of queued commands (and this is separate from the actor stash used for persistent actors during recovery or persisting).

olofwalker:
@pvlugter This is great. I'm going out on a bit of tangent below, but it does connect a bit to the billing project.

I think these numbers can be important for billing.

Spanner nodes both provide storage and processing power. It's easy to figure out how much storage a customer uses if we do Databases per customer. What's is not clear to me right now is how to figure out the computational usage incurred by a customer. I can see how different types of states can be computationally more time consuming than others.

pvlugter:
Updated: tidied up by using Akka extension for all telemetry-related things, and added some testing for the event-sourced instrumentation.

pvlugter:
This is only a start for telemetry overall, but it covers metrics for event-sourced entities and is tested now. Will merge.

PR #363, opened 2020-06-18
Upgrading to akka-grpc 1.0.0, and unfortunately this changes how Mate…

viktorklang:
…rializers are used in CS

It's hard to tell whether the changes to the Materializers might change memory retention, and I'm eliciting your opinions on the matter here.


viktorklang:
@pvlugter akka-grpc used to be Materializer-based and then it changed to be SystemMaterializer-based, which means that streams' lifecycle are no longer tied to the lifecycle of the ActorMaterializer, but the SystemMaterializer.

pvlugter:
Right, and we already switched SystemMaterializer with the Akka 2.6 upgrade, used by HTTP.

pvlugter:
So we discussed creating an issue to follow up on this, to see if we're leaking any streams when we restart everything on failures (including the retries on starting up).

PR #366, opened 2020-06-24
Added metadata support to the proxy

jroper:
This adds metadata support to the proxy, and also modifies the contract in the protocol for stateless functions.

This PR has been pulled out of the projections PR, in an effort to try and simplify that PR. Eventing support has been completely disabled in this PR (large parts of code are commented out), when the projections PR is merged, that will include all the necessary changes to eventing to make it work with the changes to everything else in this PR.  

What has been done:

* Metadata support added to protocols and proxy
* No user function support libraries have been modified. The metadata has been added to the protocol in a backwards compatible way, so they shouldn't need to be updated.
* The one exception to this is streaming support for stateless functions in JavaScript, but that needs rewriting anyway, and I have those changes ready for a future PR, so I'm not too concerned about that.

### DIFF
--- a/protocols/protocol/cloudstate/function.proto
+++ b/protocols/protocol/cloudstate/function.proto
@@ -108,7 +108,7 @@ service StatelessFunction {
     //
     // The semantics of stream closure in this protocol map 1:1 with the semantics of gRPC stream closure,
     // that is, when the client closes the stream, the stream is considered half closed, and the server
-    // should eventually, but not necessarily immediately, close the streamage with a status code and
+    // should eventually, but not necessarily immediately, close the stream with a status code and
     // trailers.
     //
     // If however the server closes the stream with a status code and trailers, the stream is immediately
--- a/protocols/protocol/cloudstate/function.proto
+++ b/protocols/protocol/cloudstate/function.proto
@@ -26,6 +26,14 @@ import "cloudstate/entity.proto";
 option java_package = "io.cloudstate.protocol";
 option go_package = "cloudstate/protocol";
 
+// A function command.
+//
+// For unary and streamed out calls, the service name, command name and payload will always be set.
+//
+// For streamed in and duplex streamed calls, the first command sent will just contain the service
+// name and command name, but no payload. This will indicate that the function has been invoked.
+// Subsequent commands on the stream will only have the payload set, the service name and command
+// name will not be set.
 message FunctionCommand {
     // The name of the service this function is on.
     string service_name = 2;
@@ -35,6 +43,9 @@ message FunctionCommand {
 
     // The command payload.
     google.protobuf.Any payload = 4;
+
+    // Metadata
+    Metadata metadata = 5;
 }
 
 message FunctionReply {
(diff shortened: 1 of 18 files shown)


protocols/protocol/cloudstate/function.proto
```diff
@@ -45,17 +56,67 @@ message FunctionReply {
         Forward forward = 3;
     }
 
-    repeated SideEffect side_effects = 4;
+    repeated SideEffect side_effects = 5;
```
viktorklang, line 59: Why 5?

proxy/core/src/main/scala/io/cloudstate/proxy/Serve.scala
```diff
@@ -206,64 +221,62 @@ object Serve {
...
+                    reply @ UserFunctionReply(Some(ClientAction(ClientAction.Action.Reply(Reply(_, metadata, _)), _)),
+                                              _,
+                                              _)
+                    ) =>
+                  (metadata.getOrElse(Metadata.defaultInstance), Source.single(reply).concat(rest))
+                case other =>
+                  (Metadata.defaultInstance, Source(other).concat(rest))
+              }
+
+              val headers = metadata.entries.collect {
```
viktorklang, line 269:
```suggestion
              val headers = metadata.entries.iterator.collect {
```

viktorklang:
@jroper Tests?

protocols/protocol/cloudstate/function.proto
```diff
@@ -45,17 +56,67 @@ message FunctionReply {
...
+    // Handle a unary command.
+    //
+    // The input command will contain the service name, command name, request metadata and the command
+    // payload. The reply may contain a direct reply, a forward or a failure, and it may contain many
+    // side effects.
     rpc handleUnary(FunctionCommand) returns (FunctionReply) {}
 
+    // Handle a streamed in command.
+    //
+    // The first message in will contain the request metadata, including the service name and command
```
sleipnir, line 73: @jroper , I have been talking about this PR with @marcellanz and we have an important question, why do you say that metadata will be sent in the first message if the metadata is defined in the protocol and is contained within the message to be forwarded? I think the metadata in that case should be passed on message by message if the underlying protocols so support. Did you mean that metadata will not be forwarded by each message? That is, will subsequent messages not contain associated metadata?

jroper:
Re: tests - without a client side implementation, it's a bit hard to test meaningfully. That will come in subsequent PRs.

protocols/protocol/cloudstate/function.proto
jroper: I'll clarify.

I think it should be this:

* If the underlying transport only supports per call metadata, that metadata will be on the first message, and not on the others.
* If the underlying transport supports per message metadata, each message will contain that metadata.

sleipnir: Thanks @jroper 
We imagined it, thank you for clarifying that. It would be nice to add this text to the PR

proxy/core/src/main/scala/io/cloudstate/proxy/HttpApi.scala
```diff
@@ -620,7 +629,11 @@ object HttpApi {
     }
   }
 
-  final def serve(services: List[(ServiceDescriptor, PartialFunction[HttpRequest, Source[ProtobufAny, NotUsed]])])(
+  final def serve(
+      services: List[
+        (ServiceDescriptor, PartialFunction[HttpRequest, Future[(List[HttpHeader], Source[ProtobufAny, NotUsed])]])
+      ]
```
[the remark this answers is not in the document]
jroper: Yeah, I think actually what would be better is a big refactor of this, not just with type aliases, but traits, so that things can be documented nicely.

PR #367, opened 2020-06-24
Metadata support for JavaScript

jroper:
Depends on #366 being merged first.

### DIFF
--- a/node-support/src/command-helper.js
+++ b/node-support/src/command-helper.js
@@ -15,16 +15,10 @@
  */
 
 const AnySupport = require("./protobuf-any");
-
-class ContextFailure extends Error {
-  constructor(msg) {
-    super(msg);
-    if (Error.captureStackTrace) {
-      Error.captureStackTrace(this, ContextFailure);
-    }
-    this.name = "ContextFailure";
-  }
-}
+const EffectSerializer = require("./effect-serializer");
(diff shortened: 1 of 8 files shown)


node-support/src/command-helper.js
```diff
@@ -49,7 +43,12 @@ module.exports = class CommandHelper {
    * @private
    */
   handleCommand(command) {
-    const ctx = this.createContext(command.id);
+    let metadata = new Metadata([]);
+    if (command.metadata && command.metadata.entries) {
```
viktorklang, line 47: Perhaps superfluous, but would it not make sense to check that the length of the entries are > 0?

protocols/protocol/cloudstate/function.proto
```diff
@@ -45,17 +56,67 @@ message FunctionReply {
         Forward forward = 3;
     }
 
-    repeated SideEffect side_effects = 4;
+    repeated SideEffect side_effects = 5;
```
viktorklang, line 59: 5?

node-support/src/command-helper.js
jroper: It will never be an empty array - it will be undefined if the array is empty, since when a sequence in protobuf is empty, nothing gets encoded, and then when protobufjs decodes that, the result is that it never sets the entries property.

PR #372, opened 2020-06-26
Add automated header settings

pvlugter:
I'm not sure what was expected here, but I keep noticing that some files have headers and some don't. The header plugin and settings are enabled and have been applied to some files, but not automatically on compile. This updates and enables automated headers across all projects and configurations.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -84,14 +84,18 @@ def akkaDiscoveryDependency(name: String, excludeThese: ExclusionRule*) =
 def akkaPersistenceCassandraDependency(name: String, excludeThese: ExclusionRule*) =
   "com.typesafe.akka" %% name % AkkaPersistenceCassandraVersion excludeAll ((excludeTheseDependencies ++ excludeThese): _*)
 
-def common: Seq[Setting[_]] = Seq(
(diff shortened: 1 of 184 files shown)


ralphlaude:
very nice @pvlugter! It is good for me.
One minor question regarding the copyright text. Shouldn't be `Copyright 2019-2020 Lightbend Inc.` instead of `Copyright 2019 Lightbend Inc.`?

pvlugter:
> One minor question regarding the copyright text. Shouldn't be `Copyright 2019-2020 Lightbend Inc.` instead of `Copyright 2019 Lightbend Inc.`?

Right, I haven't changed the existing header settings, just added the automate settings. The default for sbt-header is to just use the start year.

There are other ways to do it, although something we want to avoid is automatically applying the current year, where the first PRs of the year all end up with header changes. I think we'd want to apply the header changes separately. The akka build has a more involved setup, which we could also use here for Cloudstate:

https://github.com/akka/akka/blob/master/project/CopyrightHeader.scala

And the related issue for preserving copyright years:

https://github.com/sbt/sbt-header/issues/23

The ideal could be preserving start years and updating the end year on modifications only.

PR #378, opened 2020-07-01
Remove npm install/etc from sbt build setup

jroper:
Not sure why this is happening, but I'm pretty sure the sbt tests don't need npm install run for them. This should speed them up a little.


jroper:
The build still passes, and this has cut almost 4 minutes off the travis build time.

jroper:
GitHub is confused.

PR #384, opened 2020-07-03
Add antora extension for external code snippets

pvlugter:
This adds the `djencks/antora-aggregate-collector` extension, so we can have code snippets from examples that are outside the antora source tree.

Also add a Makefile which uses the antora docker images for building, along the lines of the Cloudflow docs.

@rstento, as discussed, a way to use those extensions for code snippets. I've just converted one example each for java and javascript docs. These extensions are unofficial and experimental currently — would be better if antora had some built-in support for something like this. It could be better to just have the code examples in the antora source tree, and add language-specific builds that compile and test. Or use the symlink approach, but it can only be a link in one direction: the original sources still need to be in antora. But this lets us include the examples from their current locations at least.

### DIFF
--- a/docs-asciidoc/.gitignore
+++ b/docs-asciidoc/.gitignore
@@ -1,0 +1,6 @@
+/.cache/
+/build/
+/node_modules/
+
+# we can't use package lock with the @djencks dependencies
+/package-lock.json
--- a/docs-asciidoc/Makefile
+++ b/docs-asciidoc/Makefile
@@ -1,0 +1,53 @@
+# Make Cloudstate documentation
+
+antora_version      := 2.3.3
+antora_docker_image := antora/antora
+
+docs_dir  := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+root_dir  := ${docs_dir}/..
+build_dir := ${docs_dir}/build
+site_dir  := ${build_dir}/site
+base_dir  := /antora/$(shell basename ${docs_dir})
+
+build: clean install html print-site
+
+clean:
+	rm -rf ${build_dir}
+
+# note: we can't currently use package lock with the @djencks dependencies
+install:
+	docker run \
+		-u $(shell id -u):$(shell id -g) \
+		-v ${root_dir}:/antora \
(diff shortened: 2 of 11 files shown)


rstento:
@pvlugter, another alternative is to include the sample code from the samples repos, such as cloudstate/samples-js-shoppingcart. It seems more straightforward in these smaller repos to set up a samples tree structure to be includeable by Antora. I'll forward you an email I sent to Nash and Ed where I changed the directory structure a bit. Then, these code snippets could be pulled in by both the OSS and the commercial doc.

pvlugter:
Contributors have already discussed not having the samples in separate repos—more difficult to maintain and update when separate—so we'll likely move the samples into the language support repos. But having sample code under the antora source structure seems fine—since antora is inflexible on this, while it should be possible to configure build tools for an unconventional location.

vkorenev:
I am not sure that putting source files in unconventional locations in samples is a good solution.
Those trying to learn how to create a Cloudstate service by looking at a sample will notice that the source files are not placed in the default location, which will inevitably raise a question why. Even if the sample docs explain that this unusual layout is needed for Antora, this will create additional cognitive load on a person who is new to Cloudstate.
Also, some people might want to use a sample as a template for their own project by copying it and then changing source files. They will end up with unconventional project structure without a good reason.
So I think it is better to keep samples structure simple at the expense of more complex doc project rather than vice versa.

pvlugter:
@vkorenev, good point, and the changes in this PR allow for the source files to be pulled in to the docs source tree.

We've also been discussing that samples could be published as zip files to download, rather than cloning the language support repos, and we could then make sure that it's just the sample in a conventional structure. The example snippets also require tag comments in the sample files, and it would be nice to strip these out too. The samples are used for the doc snippets, the TCK app tests, and as runnable samples, and it would be useful to have a setup that supports all of these.

So what we're thinking: having the samples and docs in with the language supports would be good — means that changes to code, docs, and samples can all happen together and these always stay aligned. We can look at publishing zip downloads for the samples (on each release) which will be clean and properly structured.

Issue #375, opened 2020-06-26
Event sourcing – sequence of handling emitted events (JavaScript- vs. Java-Support)


marcellanz:
The documentation for the Java [Event Sourcing Support states](https://cloudstate.io/docs/core/current/user/lang/java/eventsourced.html#emitting-events) that:

>Invoking emit will immediately invoke the associated event handler for that event - this both validates that the event can be applied to the current state, as well as updates the state so that subsequent processing in the command handler can use it.

and this seems to be implemented [that way](https://github.com/cloudstateio/cloudstate/blob/v0.5.1/java-support/src/main/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImpl.scala#L227-L234).

For [JavaScript](https://cloudstate.io/docs/core/current/user/lang/javascript/eventsourced.html#emitting-events), it is not defined by the documentation what the sequence of emitting events and the handling of emitted events is. The JavaScript  implementation seems [not to follow the same sequence](https://github.com/cloudstateio/cloudstate/blob/v0.5.1/node-support/src/eventsourced-support.js#L122-L128) as the one of the Java Support I think. The JavaScript implementation instead, collects all emitted events while running the command handler and then, after the command was handled, [the collected events get handled as a whole](https://github.com/cloudstateio/cloudstate/blob/v0.5.1/node-support/src/eventsourced-support.js#L132-L140).

If I understood it right, I think it is relevant for both, the error behaviour as well as what the entity state would be after a call to `emit()` in the command handler; or further calls to `emit()` within the same handler.

@jroper, I would assume both implementations having the same behaviour, wdyt? 

(I tried to validate this with a unit-test for JavaScript. I started, but after a while stopped as it, for me, was not that simple to do. I think there are no tests like the ones by `crdt-handler-test.js`. There is `js-shopping-cart/test/test.js` but they have a different scope.)

jroper:
I think we should rewrite the event sourced support for JavaScript. I don't like it. We actually have someone working on a redux like API at the moment, so let's see what comes of that.

marcellanz:
I think with #392 we can close this one. wdyt @jroper @pvlugter ?
Although we should change the documentation about command handler can use the local state in subsequent processing.

marcellanz:
At the same time, I find the ability to loose local changed state for subsequent processing a bit of a drawback; as it is a feature that is lost because of a condition that usually happens sparingly.

pvlugter:
[...]

PR #392, opened 2020-07-14
Handle failures in user functions

### LINKED ISSUE #333: [Node support] Exceptions occurring in user code are not handled properly
Right now an exception thrown in a command handler causes termination of node process:
```
/opt/app/build/Api.js:35
            throw new Error("error");
            ^

Error: error
    at Object.Api.getShoppingCart [as GetShoppingCart] (/opt/app/build/Api.js:35:19)
    at StatelessServices.handleUnary (/opt/app/node_modules/cloudstate/src/stateless-support.js:158:68)
    at /opt/app/node_modules/grpc/src/server.js:593:13
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! shopping-service@1.0.0 start-no-prestart: `node --inspect=0.0.0.0 build/index.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the shopping-service@1.0.0 start-no-prestart script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /root/.npm/_logs/2020-05-22T20_14_15_268Z-debug.log
```
Calling `ctx.fail(err)` is not handled properly as well:
```
$ grpcurl -plaintext -d '{ "user_id": "vk", "product_id": "apple" }' localhost:9000 com.example.shoppingcart.ShoppingCart/RemoveItem
Error invoking method "com.example.shoppingcart.ShoppingCart/RemoveItem": grpc call for "com.example.shoppingcart.ShoppingCart.RemoveItem" failed: EOF
```
Java Support library has the same problem: #344.

### LINKED ISSUE #344: [Java support] Exception in command handler results in malformed gRPC response
If an exception is thrown in a command handler of a user function written in Java (including `FailInvoked` thrown by `CommandContext.fail()`, the caller (gRPC client) does not receive a gRPC error. Instead the gRPC response seems to be truncated.

Output from `grpcurl` client:
```
Error invoking method "com.example.inventory.ProductInventory/ReserveProduct": grpc call for "com.example.inventory.ProductInventory.ReserveProduct" failed: EOF
```

pvlugter:
Resolves #344 and resolves #333. Still to do:

- [x] add tests for the failure paths

Have only manually checked this, will follow up with the tests.

<details>
<summary>Old behaviour - output and proxy logs</summary>

* gRPC request for command using `context.fail`:

  ```
  ❯ grpcurl -plaintext -d '{"user_id": "test", "product_id": "abc"}' localhost:9000 com.example.shoppingcart.ShoppingCart/RemoveItem
  Error invoking method "com.example.shoppingcart.ShoppingCart/RemoveItem": grpc call for "com.example.shoppingcart.ShoppingCart.RemoveItem" failed: EOF
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Cannot remove item abc because it is not in the cart.
  ```

* HTTP request for command using `context.fail`:

  ```
  ❯ http POST :9000/cart/test/items/abc/remove
  HTTP/1.1 500 Internal Server Error
  Connection: close
  Content-Length: 0
  Server: akka-http/10.1.12
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Cannot remove item abc because it is not in the cart.
  ERROR akka.actor.ActorSystemImpl - Internal server error, sending 500 response
  java.util.NoSuchElementException: head of empty stream
   at akka.stream.scaladsl.Sink$.$anonfun$head$3(Sink.scala:198)
   [...]
  INFO akka.actor.ActorSystemImpl - Request timeout encountered for request [POST /cart/test/items/abc/remove Empty]
  ```

* gRPC request for command with exception:

  ```
  ❯ grpcurl -plaintext -d '{"user_id": "test"}' localhost:9000 com.example.shoppingcart.ShoppingCart/GetCart
  Error invoking method "com.example.shoppingcart.ShoppingCart/GetCart": grpc call for "com.example.shoppingcart.ShoppingCart.GetCart" failed: EOF
  ```
  Proxy logs:
  ```
  ERROR akka.actor.OneForOneStrategy - INTERNAL
  io.grpc.StatusRuntimeException: INTERNAL
   at io.grpc.Status.asRuntimeException(Status.java:533)
   at akka.grpc.internal.AkkaNettyGrpcClientGraphStage$$anon$1.onCallClosed(AkkaNettyGrpcClientGraphStage.scala:165)
   [...]
  WARN io.cloudstate.proxy.eventsourced.EventSourcedEntity - Stopped but we have a current action id 2:test:1
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Unexpected entity termination
  ```

* HTTP request for command with exception:

  ```
  ❯ http :9000/carts/test
  HTTP/1.1 500 Internal Server Error
  Connection: close
  Content-Length: 0
  Server: akka-http/10.1.12
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Unexpected entity termination
  ERROR akka.actor.ActorSystemImpl - Internal server error, sending 500 response
  java.util.NoSuchElementException: head of empty stream
   at akka.stream.scaladsl.Sink$.$anonfun$head$3(Sink.scala:198)
   [...]
  INFO akka.actor.ActorSystemImpl - Request timeout encountered for request [GET /carts/test Empty]
  ```

</details>

There's already some currently unused code for `Serve.CommandException` and `HttpEndpoint.transformFailedRequest`. Add simple fixes: use the `CommandException` on failures (which already gets converted to gRPC errors), add a default catch for exceptions in `HttpApi`, and add a default catch for exceptions in java-support event sourced and report as errors.

<details>
<summary>New behaviour - output and proxy logs</summary>

* gRPC request for command using `context.fail`:

  ```
  ❯ grpcurl -plaintext -d '{"user_id": "test", "product_id": "abc"}' localhost:9000 com.example.shoppingcart.ShoppingCart/RemoveItem
  ERROR:
    Code: Unknown
    Message: Cannot remove item abc because it is not in the cart.
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Cannot remove item abc because it is not in the cart.
  ```

* HTTP request for command using `context.fail`:

  ```
  ❯ http POST :9000/cart/test/items/abc/remove
  HTTP/1.1 500 Internal Server Error
  Content-Length: 53
  Content-Type: text/plain; charset=UTF-8
  Server: akka-http/10.1.12
  
  Cannot remove item abc because it is not in the cart.
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Cannot remove item abc because it is not in the cart.
  ```

* gRPC request for command with exception:

  ```
  ❯ grpcurl -plaintext -d '{"user_id": "test"}' localhost:9000 com.example.shoppingcart.ShoppingCart/GetCart
  ERROR:
    Code: Unknown
    Message: Unexpected failure: boom!
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Unexpected failure: boom!
  ```

* HTTP request for command with exception:

  ```
  ❯ http :9000/carts/test
  HTTP/1.1 500 Internal Server Error
  Content-Length: 25
  Content-Type: text/plain; charset=UTF-8
  Server: akka-http/10.1.12

  Unexpected failure: boom!
  ```
  Proxy logs:
  ```
  ERROR io.cloudstate.proxy.Serve$ - User Function responded with a failure: Unexpected failure: boom!
  ```

</details>

### DIFF
--- a/java-support/src/main/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImpl.scala
+++ b/java-support/src/main/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImpl.scala
@@ -20,6 +20,7 @@ import java.util.Optional
 
 import akka.NotUsed
 import akka.actor.ActorSystem
+import akka.event.{Logging, LoggingAdapter}
 import akka.stream.scaladsl.Flow
 import com.google.protobuf.{Descriptors, Any => JavaPbAny}
 import com.google.protobuf.any.{Any => ScalaPbAny}
@@ -35,14 +36,16 @@ import io.cloudstate.javasupport.impl.{
   ResolvedEntityFactory,
   ResolvedServiceMethod
 }
+import io.cloudstate.protocol.entity.{Command, Failure}
 import io.cloudstate.protocol.event_sourced.EventSourcedStreamIn.Message.{
   Command => InCommand,
   Empty => InEmpty,
   Event => InEvent,
   Init => InInit
 }
-import io.cloudstate.protocol.event_sourced.EventSourcedStreamOut.Message.{Reply => OutReply}
+import io.cloudstate.protocol.event_sourced.EventSourcedStreamOut.Message.{Failure => OutFailure, Reply => OutReply}
 import io.cloudstate.protocol.event_sourced._
+import scala.util.control.NonFatal
 
 final class EventSourcedStatefulService(val factory: EventSourcedEntityFactory,
                                         override val descriptor: Descriptors.ServiceDescriptor,
@@ -65,11 +68,53 @@ final class EventSourcedStatefulService(val factory: EventSourcedEntityFactory,
       this
 }
 
+object EventSourcedImpl {
+  final case class EntityException(entityId: String, commandId: Long, commandName: String, message: String)
+      extends RuntimeException(message)
+
+  object EntityException {
+    def apply(message: String): EntityException =
+      EntityException(entityId = "", commandId = 0, commandName = "", message)
+
+    def apply(command: Command, message: String): EntityException =
+      EntityException(command.entityId, command.id, command.name, message)
+
+    def apply(context: CommandContext, message: String): EntityException =
+      EntityException(context.entityId, context.commandId, context.commandName, message)
+  }
+
+  object ProtocolException {
+    def apply(message: String): EntityException =
+      EntityException(entityId = "", commandId = 0, commandName = "", "Protocol error: " + message)
+
+    def apply(init: EventSourcedInit, message: String): EntityException =
+      EntityException(init.entityId, commandId = 0, commandName = "", "Protocol error: " + message)
+
+    def apply(command: Command, message: String): EntityException =
+      EntityException(command.entityId, command.id, command.name, "Protocol error: " + message)
+  }
+
+  def failure(cause: Throwable): Failure = cause match {
+    case e: EntityException => Failure(e.commandId, e.message)
+    case e => Failure(description = "Unexpected failure: " + e.getMessage)
+  }
+
+  def failureMessage(cause: Throwable): String = cause match {
+    case EntityException(entityId, commandId, commandName, _) =>
+      val commandDescription = if (commandId != 0) s" for command [$commandName]" else ""
+      val entityDescription = if (entityId.nonEmpty) s"entity [$entityId]" else "entity"
+      s"Terminating $entityDescription due to unexpected failure$commandDescription"
+    case _ => "Terminating entity due to unexpected failure"
+  }
+}
+
 final class EventSourcedImpl(_system: ActorSystem,
                              _services: Map[String, EventSourcedStatefulService],
                              rootContext: Context,
                              configuration: Configuration)
     extends EventSourced {
+  import EventSourcedImpl._
+
   private final val system = _system
   private final val services = _services.iterator
     .map({
@@ -78,6 +123,8 @@ final class EventSourcedImpl(_system: ActorSystem,
         (name, if (esss.snapshotEvery == 0) esss.withSnapshotEvery(configuration.snapshotEvery) else esss)
     })
     .toMap
+
+  private val log = Logging(system.eventStream, this.getClass)
 
   /**
    * The stream. One stream will be established per active entity.
@@ -99,18 +146,17 @@ final class EventSourcedImpl(_system: ActorSystem,
         case (Seq(EventSourcedStreamIn(InInit(init), _)), source) =>
           source.via(runEntity(init))
         case _ =>
-          // todo better error
-          throw new RuntimeException("Expected Init message")
+          throw ProtocolException("Expected Init message")
       }
       .recover {
-        case e =>
-          // FIXME translate to failure message
-          throw e
+        case error =>
+          log.error(error, failureMessage(error))
+          EventSourcedStreamOut(OutFailure(failure(error)))
       }
 
   private def runEntity(init: EventSourcedInit): Flow[EventSourcedStreamIn, EventSourcedStreamOut, NotUsed] = {
     val service =
-      services.getOrElse(init.serviceName, throw new RuntimeException(s"Service not found: ${init.serviceName}"))
+      services.getOrElse(init.serviceName, throw ProtocolException(init, s"Service not found: ${init.serviceName}"))
     val handler = service.factory.create(new EventSourcedContextImpl(init.entityId))
     val thisEntityId = init.entityId
 
@@ -137,22 +183,18 @@ final class EventSourcedImpl(_system: ActorSystem,
           (event.sequence, None)
         case ((sequence, _), InCommand(command)) =>
           if (thisEntityId != command.entityId)
-            throw new IllegalStateException("Receiving entity is not the intended recipient of command")
-          val cmd = ScalaPbAny.toJavaProto(command.payload.get)
-          val context = new CommandContextImpl(thisEntityId,
-                                               sequence,
-                                               command.name,
-                                               command.id,
-                                               service.anySupport,
-                                               handler,
-                                               service.snapshotEvery)
+            throw ProtocolException(command, "Receiving entity is not the intended recipient of command")
+          val cmd =
+            ScalaPbAny.toJavaProto(command.payload.getOrElse(throw ProtocolException(command, "No command payload")))
+          val context =
+            new CommandContextImpl(thisEntityId, sequence, command.name, command.id, service.anySupport, log)
 
           val reply = try {
-            handler.handleCommand(cmd, context) // FIXME is this allowed to throw
+            handler.handleCommand(cmd, context)
           } catch {
-            case FailInvoked =>
-              Optional.empty[JavaPbAny]()
-            // Ignore, error already captured
+            case FailInvoked => Optional.empty[JavaPbAny]() // Ignore, error already captured
+            case e: EntityException => throw e
+            case NonFatal(error) => throw EntityException(command, "Unexpected failure: " + error.getMessage)
           } finally {
             context.deactivate() // Very important!
           }
@@ -160,10 +202,16 @@ final class EventSourcedImpl(_system: ActorSystem,
           val clientAction = context.createClientAction(reply, false)
 
           if (!context.hasError) {
-            val endSequenceNumber = sequence + context.events.size
+            // apply events from successful command to local entity state
+            context.events.zipWithIndex.foreach {
+              case (event, i) =>
+                handler.handleEvent(ScalaPbAny.toJavaProto(event), new EventContextImpl(thisEntityId, sequence + i + 1))
+            }
 
+            val endSequenceNumber = sequence + context.events.size
+            val performSnapshot = (endSequenceNumber / service.snapshotEvery) > (sequence / service.snapshotEvery)
             val snapshot =
-              if (context.performSnapshot) {
+              if (performSnapshot) {
                 val s = handler.snapshot(new SnapshotContext with AbstractContext {
                   override def entityId: String = entityId
                   override def sequenceNumber: Long = endSequenceNumber
@@ -195,9 +243,9 @@ final class EventSourcedImpl(_system: ActorSystem,
              ))
           }
         case (_, InInit(i)) =>
-          throw new IllegalStateException("Entity already inited")
+          throw ProtocolException(init, "Entity already inited")
         case (_, InEmpty) =>
-          throw new IllegalStateException("Received empty/unknown message")
+          throw ProtocolException(init, "Received empty/unknown message")
       }
       .collect {
         case (_, Some(message)) => EventSourcedStreamOut(message)
@@ -213,8 +261,7 @@ final class EventSourcedImpl(_system: ActorSystem,
                            override val commandName: String,
                            override val commandId: Long,
                            val anySupport: AnySupport,
-                           val handler: EventSourcedEntityHandler,
-                           val snapshotEvery: Int)
+                           val log: LoggingAdapter)
       extends CommandContext
       with AbstractContext
       with AbstractClientActionContext
@@ -222,16 +269,14 @@ final class EventSourcedImpl(_system: ActorSystem,
       with ActivatableContext {
 
     final var events: Vector[ScalaPbAny] = Vector.empty
-    final var performSnapshot: Boolean = false
 
     override def emit(event: AnyRef): Unit = {
       checkActive()
-      val encoded = anySupport.encodeScala(event)
-      val nextSequenceNumber = sequenceNumber + events.size + 1
-      handler.handleEvent(ScalaPbAny.toJavaProto(encoded), new EventContextImpl(entityId, nextSequenceNumber))
-      events :+= encoded
-      performSnapshot = (snapshotEvery > 0) && (performSnapshot || (nextSequenceNumber % snapshotEvery == 0))
+      events :+= anySupport.encodeScala(event)
     }
+
+    override protected def logError(message: String): Unit =
+      log.error("Fail invoked for command [{}] for entity [{}]: {}", commandName, entityId, message)
   }
 
   class EventSourcedContextImpl(override final val entityId: String) extends EventSourcedContext with AbstractContext
--- a/java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
+++ b/java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
@@ -1,0 +1,305 @@
+/*
+ * Copyright 2019 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudstate.javasupport.impl.eventsourced
+
+import com.google.protobuf.Empty
+import io.cloudstate.javasupport.EntityId
+import io.cloudstate.javasupport.eventsourced._
+import io.cloudstate.testkit.eventsourced.{EventSourcedMessages, TestEventSourcedClient}
+import org.scalatest.{BeforeAndAfterAll, Matchers, WordSpec}
+import scala.collection.mutable
+import scala.reflect.ClassTag
+
+class EventSourcedImplSpec extends WordSpec with Matchers with BeforeAndAfterAll {
+  import EventSourcedImplSpec._
+  import EventSourcedMessages._
+  import ShoppingCart.Item
+  import ShoppingCart.Protocol._
+
+  val service: TestEventSourcedService = ShoppingCart.testService
+  val client: TestEventSourcedClient = TestEventSourcedClient(service.port)
+
+  override def afterAll(): Unit = {
+    client.terminate()
+    service.terminate()
+  }
+
+  "EventSourcedImpl" should {
+
+    "manage entities with expected commands and events" in {
+      val entity = client.connect
+      entity.send(init(ShoppingCart.Name, "cart"))
+      entity.send(command(1, "cart", "GetCart"))
+      entity.expect(reply(1, EmptyCart))
+      entity.send(command(2, "cart", "AddItem", addItem("abc", "apple", 1)))
+      entity.expect(reply(2, EmptyPayload, itemAdded("abc", "apple", 1)))
+      entity.send(command(3, "cart", "AddItem", addItem("abc", "apple", 2)))
+      entity.expect(reply(3, EmptyPayload, Seq(itemAdded("abc", "apple", 2)), cartSnapshot(Item("abc", "apple", 3))))
+      entity.send(command(4, "cart", "GetCart"))
+      entity.expect(reply(4, cart(Item("abc", "apple", 3))))
+      entity.send(command(5, "cart", "AddItem", addItem("123", "banana", 4)))
+      entity.expect(reply(5, EmptyPayload, itemAdded("123", "banana", 4)))
+      entity.passivate()
+      val reactivated = client.connect
+      reactivated.send(init(ShoppingCart.Name, "cart", snapshot(3, cartSnapshot(Item("abc", "apple", 3)))))
+      reactivated.send(event(4, itemAdded("123", "banana", 4)))
+      reactivated.send(command(1, "cart", "GetCart"))
+      reactivated.expect(reply(1, cart(Item("abc", "apple", 3), Item("123", "banana", 4))))
+      reactivated.passivate()
+    }
+
+    "fail when first message is not init" in {
+      service.expectLogError("Terminating entity due to unexpected failure") {
+        val entity = client.connect
+        entity.send(command(1, "cart", "command"))
+        entity.expect(failure("Protocol error: Expected Init message"))
+        entity.expectClosed()
+      }
+    }
+
+    "fail when service doesn't exist" in {
+      service.expectLogError("Terminating entity [foo] due to unexpected failure") {
+        val entity = client.connect
+        entity.send(init(serviceName = "DoesNotExist", entityId = "foo"))
+        entity.expect(failure("Protocol error: Service not found: DoesNotExist"))
+        entity.expectClosed()
+      }
+    }
+
+    "fail when command payload is missing" in {
+      service.expectLogError("Terminating entity [cart] due to unexpected failure for command [foo]") {
+        val entity = client.connect
+        entity.send(init(ShoppingCart.Name, "cart"))
+        entity.send(command(1, "cart", "foo", payload = None))
+        entity.expect(failure(1, "Protocol error: No command payload"))
+        entity.expectClosed()
+      }
+    }
+
+    "fail when command entity id is incorrect" in {
+      service.expectLogError("Terminating entity [cart2] due to unexpected failure for command [foo]") {
+        val entity = client.connect
+        entity.send(init(ShoppingCart.Name, "cart1"))
+        entity.send(command(1, "cart2", "foo"))
+        entity.expect(failure(1, "Protocol error: Receiving entity is not the intended recipient of command"))
+        entity.expectClosed()
+      }
+    }
+
+    "fail when entity is sent multiple init" in {
+      service.expectLogError("Terminating entity [cart] due to unexpected failure") {
+        val entity = client.connect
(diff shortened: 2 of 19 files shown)


pvlugter:
I've added some tests for this PR now. Starting to create a simple testkit, with probes and send/expect methods, for testing either side of the protocol — just event sourced entities currently for these test cases.

java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
```diff
@@ -0,0 +1,211 @@
...
+      val expectedMessage = s"No command handler found for command [foo] on class ${ShoppingCart.TestCartClassName}"
+      entity.expect(failure(1, s"Unexpected failure: $expectedMessage"))
+      entity.passivate()
+    }
+
+    "fail when command uses context fail" in {
+      val entity = client.connect
+      entity.send(init(ShoppingCart.Name, "cart"))
+      entity.send(command(1, "cart", "AddItem", addItem("foo", "bar", -1)))
+      entity.expect(failure(1, "Cannot add negative quantity of item [foo]"))
```
marcellanz, line 105: the current TCK still mentions:
>// FIXME Expects entity.Failure, but gets lientAction.Action.Failure(Failure(commandId, msg)))

here: https://github.com/cloudstateio/cloudstate/blob/v0.5.1/tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala#L230

And I'm curious if we should expect an `actionFailure` for this kind of error, where the user function signals an error through `context.Fail()`.

I remember the TCK being inconsistent in this regard I think and the current Go implementation still sends a `client action failure`, even if a failure would be appropriate.

pvlugter: Good point. Which level to return failures should be looked at more closely — with either `EventSourcedStreamOut.Message.Failure` and `ClientAction.Action.Failure`.

I was initially looking to just make sure errors are reported, rather than the EOF in the current behaviour. But it would make sense to make sure the protocol is sound now too.

`Context.fail` has the same behaviour as before — returns a client action failure. The testkit messages currently always assume a client action failure — I'll update to distinguish these failures.

For the unexpected exceptions in handlers it was easy to have it also be a context error and report in the same way, with a client action failure — this should probably be an event sourced message failure instead. But it seems that the current behaviour of the proxy, however, is to translate these failures into client action failures anyway, and continuing with processing rather than cancelling the stream.

https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala#L332-L337

https://github.com/cloudstateio/cloudstate/blob/master/proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala#L266-L269

If any other exceptions are thrown (outside of the handler, for protocol errors, service not found, wrong recipient, or so on), I've updated this to throw a gRPC exception when the stream is stopping, so that it gets reported while cancelling — but this could also be an explicit event sourced message failure instead of a stream error, and there's a previous fixme suggesting this.

So currently, the event sourced message failure is not expected at all, it's either a client action failure or it's the stream completing with an error. At least in the java support. And if event sourced message failures are returned, they are not currently treated differently in the proxy, but translated into client action failures.

Thanks for raising this @marcellanz. I see you've also asked about failures in the draft spec. I can look at updating things to be in line with the spec discussions — where client action failure is only used to signal an explicit context.fail, and event sourced message failures are used for all other exceptions and these are treated as fatal on the proxy side.

pvlugter: I've updated this PR to be aligned with discussions — java-support only uses client action failure for explicit context.fail and all other exceptions are reported as entity failures, and the proxy now treats entity failures as fatal and doesn't leak specific details of these failures to the client.

marcellanz: Thanks for looking at updating things to be in line with the spec discussions @pvlugter.

Probably out of the scope of this issue/PR comment, I'm interested in to define what is expected from a user entity after issuing fatal and non-fatal failures in the user entity now handled by the proxy differently in terms of expectations of state consistency of an entities local state.

This would not change anything changed with this PR I think, so I'll add my question to the draft spec.

java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
```diff
@@ -0,0 +1,267 @@
...
+      entity.expect(failure(1, s"No command handler found for command [foo] on ${ShoppingCart.TestCartClass}"))
+      entity.expectClosed()
+    }
+
+    "fail action when command handler uses context fail" in {
+      val entity = client.connect
+      entity.send(init(ShoppingCart.Name, "cart"))
+      entity.send(command(1, "cart", "AddItem", addItem("foo", "bar", -1)))
+      entity.expect(actionFailure(1, "Cannot add negative quantity of item [foo]"))
+      entity.passivate()
```
marcellanz, line 158: Does the proxy passivate an entity on the reception of an action failure? If, yes, this would answer a question of mine regarding the user entities state, after such a failure.

java-support/src/main/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImpl.scala
```diff
@@ -99,18 +123,17 @@ final class EventSourcedImpl(_system: ActorSystem,
...
-          // todo better error
-          throw new RuntimeException("Expected Init message")
+          throw ProtocolException("Expected Init message")
       }
       .recover {
-        case e =>
-          // FIXME translate to failure message
-          throw e
+        case EntityException(id, message) => failure(id, message)
+        case ProtocolException(id, message) => failure(id, "Protocol error: " + message)
```
marcellanz, line 130: Both cases produce a `EventSourcedStreamOut.OutFailure.Failure` is that correct?

Don't we send:

`EntityException` => `EventSourcedStreamOut.Message.Failure`
`ProtocolException`  => `ClientAction.Action.Failure`

marcellanz: ^^^ edited! got it wrong at first :)

java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
pvlugter: The proxy sees client action failures as an expected reply with a deliberate failure. It processes it the same as non-failed replies (even including persisting events and snapshots if they are included). The entity is not passivated, and the stream is not closed, for action failures. The examples use context.fail for validation, and I assume the expectation is for the entity to remain in a good state.

The `passivate` here in this test is just to end the interaction, while other tests have `expectClosed` if it's an entity failure plus stream closed on the service side — this is testing the java-support, while the proxy behaviour is tested in the exception handling spec.

java-support/src/main/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImpl.scala
pvlugter: Any exceptions at this point (in the `recover` for the akka stream) are going to be converted into entity failures — these are all unexpected, while `context.fail` in the java-support throws a `FailInvoked` which is caught when invoking handlers and becomes a failure in the reply. Other exceptions thrown anywhere in the akka stream will fail the stage they occur in and then cancel (upstream) and complete (downstream) with the failure. The `recover` allows the exception to be converted into a final message before the stream is closed.

So the behaviour now is that any unexpected exception (not `FailInvoked`) in the service creates an entity failure message plus closes the stream, and on the proxy side the failure is reported to the client with a generic message and the entity is stopped and stream closed.

I distinguished the protocol exception — some failure in the protocol interaction with the proxy — from exceptions in the entity. The only differences in the end are that the entity exception message is left as is for the failure description, the protocol exception has "Protocol error:" and any other exceptions have "Unexpected failure:" — but maybe it's more understandable in the impl to just have one exception type for entity failures. The special exception type allows the command id to be added when available, which is its main purpose.

java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
pvlugter: Once we're clear on the expected behaviour, we should express all of this in the TCK tests as well, and the specification.

sleipnir: I agree

java-support/src/test/scala/io/cloudstate/javasupport/impl/eventsourced/EventSourcedImplSpec.scala
pvlugter: On the behaviour and currently implemented expectations:

* Action failures in replies are handled by the proxy in the same way as non-failure replies (same as it was before this PR). So the user entity is expected to be in a consistent state after explicit calls to `context.fail` — it can be interpreted as reply with failure message, and usually the state is unchanged (no events).

* Entity failures now error and stop the proxy entity actor and close the stream (this has changed behaviour in this PR, previously it continued processing). The user entity should be cleared and the stream closed for these unexpected failures.

I'm assuming these are aligned with the original intent and spec discussions now.

But what I'm hearing in your questions: should any failure, including `context.fail`, be considered fatal for the current state of the entity. So that a user can call `context.fail` anywhere that they think things have become inconsistent. I can see how that makes sense, especially with it being a `fail` call. In which case, I think we would want a different mechanism for returning status codes and descriptions for client request errors. Currently `context.fail` means reply with a client request error (immediately) but continue with this active entity otherwise, and to fail the entity fully an exception can be thrown.

sleipnir: The fact is not very clear if the user wants to interrupt the flow due to a fatal error or just inform that an expectation has failed and he is only intending to inform the client.
Perhaps the context, as well as the protocol, allows to choose the type of failure or the desired return, perhaps this can be implemented with the two types of failure already existing. In Rust, you can wait when we want to report that an expectation has not been met, perhaps the context can prove the same type of mechanism, namely context.fail () for a catastrophic failure and context.expect () to just report, the first one finishes the flow trying to keep the state consistent while the second one follows the normal flow without closing it, probably both without events

marcellanz: Exactly, this is what I thought about.

As soon as a command handler emits events, and the events get applied immediately to the local state, a later call to `context.fail` would leave the entities local state in an inconsistent state the proxy is not aware of I think. The entity could send the emitted events back to the proxy, `EventSourcedReply` would allow this, but this would leave the entity in an awkward semi-complete state I think. The option then could be, that the user entity informs the proxy about the `context.fail`-error with an action failure, but also lets the proxy re-establish its state; the proxy in this situation knows best about the entities consistent state.

Saying this, if the client sends the same command, the user entity will probably end in the same situation and again will fail as described with the context failed. But if the client sends a command with different input, because the client was informed through the client action failure, the entity could proceed.

sleipnir: @marcellanz What you describe sounds like idempotency. That would be very desirable. It seems to me that context.fail would be called in commands never in events, which alone would guarantee that the state would be consistent, since events have not yet been issued.

marcellanz: >It seems to me that context.fail would be called in commands never in events

Correct, should. But the following can happen:

- start to handle command
- **`call context.emit()`** which emit events and get them **immediately applied** as by current documentation [described](https://cloudstate.io/docs/core/current/user/lang/java/eventsourced.html#emitting-events) or see below 
(note: this is not done as documented in the JS implementation currently, see https://github.com/cloudstateio/cloudstate/issues/375)
- **`call context.fail()`**
- the user even can call again context.emit(), but not `context.fail()`

We even say:
>this both validates that the event can be applied to the current state, as well as updates the state so that subsequent processing in the command handler can use it.

so this is intended to happen.


**emitting-events**
https://cloudstate.io/docs/core/current/user/lang/java/eventsourced.html#emitting-events states:
**Invoking emit will immediately invoke the associated event handler** for that event - this both validates that the event can be applied to the current state, as well as updates the state so that subsequent processing in the command handler can use it.

marcellanz: As we say we'd "**validate that the event can be applied to the current state**", implies that we, if we fail, are inconsistent in our local state. If we don't allow `context.fail` to be called after `context.emit()` every failure in `context.emit()` or exception after `context.emit()`, would lead to a fatal, stream closing, failure.

sleipnir: Yes, I agree, but this is because in the current scenario it seems to be assumed that this is not really a real failure. I think that one should first separate the concepts of failure and validation and then define the best chain of events to be carried out.
For me it would be natural to call context.emit (), context.expect (), context.emit () because I am just reporting something, but against intuitive to call context.emit (), context.fail (), context.emit because it seems that fail now should break the whole thing and interrupt the flow of events until my last issue () before fail

marcellanz: I see. I think this is quite a burden for the user to ensure doing the right thing at the right time.

There is something else, that we describe in [terminology](https://cloudstate.io/docs/core/current/user/features/eventsourced.html) – see below.

During command handling, we are not allowed to update the local state, but we also describe that we do exactly this, because we interweave the span of the command handling with the update of the local state right within `context.emit`.

One understanding the relation of commands and events, events are side effects happened after a command has completed as a whole I think. If the command did fail, there should be no events, and there are none after a `context.fail` because the proxy does not know about them. A further command would emit, based on inconsistent local state, inconsistent new events and if this new command succeeds, the proxy would get tainted events from the entity I think.

**Command handler**
>A command handler is the code that handles a command. It may validate the command using the current state, and may emit events as part of its processing. A command handler **must not update the state of the entity directly**, only indirectly by emitting events. If a command handler does update the state, then when the entity is passivated (removed from memory), those updates will be lost.

marcellanz: I see this might all be, we should assume it does not happen, but if it does, it corrupts state and is not easily detected. 

The Client Action Failure is just a normal hint for the client something did not go well, like the negative added products in the shopping cart, it might be just a hint for the user to "try" again and is not exceptional. Especially in this case, if the user retries, failed validation in the command should not lead to corrupted state.

sleipnir: There are flaws and flaws in how the user informs about proxy or that each one must define the result of the flow of events.

pvlugter: Agree, we need to define this clearly. And make it as intuitive as possible for users.

With the scenario of emit and then fail, the java-support implementation will actually have entities in inconsistent states. Events are applied immediately on emit, as described, but no events are returned in the reply for action failures. So after an emit-then-fail a subsequent get on the entity will return an updated state, but once the entity is passivated and reactivated the state will not have those events. Users can't call emit after fail though (as fail throws an exception in the java-support and node-support implementations). So the current implementation does have the possible bug suggested by failing after immediately applied emits. It needs to either respond with any events emitted and applied before a fail so these can be persisted, or wait with applying events until after successful handling of a command (like the node-support implementation does).

It's useful to have the emitted events validated against the event handlers, before responding, but we need to avoid this being in conflict with a likely expectation for handlers to be atomic — all events or none applied — and that `fail` is probably understood as failing the whole handler, including any events emitted. The usual approach is to wait for events to be persisted successfully before updating the in-memory state, but here that would require additional communication from the proxy, but the entity could still have atomicity for handlers by applying after command processing. So I'm in favour of the node-support version being the expected behaviour.

I'm not sure that we would need multiple ways to fail, as users should only need to fail an invalid command, and not the entity as a whole. Only valid commands should be applied, with the state kept consistent. The current `context.fail` as meaning fail the current command invocation, with no events applied, makes sense to me.

I've extended this test to show the incorrect behaviour. And I think the change we should make now is having the java-support delay applying events until after successful command processing.

marcellanz: Thanks @pvlugter for the summary, I agree.

marcellanz: An additonal thought:
The entity actually knows how to reset its local state to the state it had before the command handler started through a snapshot it would take and apply to itself, or hold the last snapshot plus events received from its initial state. This would prevent any communication from the proxy to restore its local state.

sleipnir: @marcellanz  thinks this is complicated, it would solve the problem but it breaks the basic premise that it is the proxy that deals with the state. Guys, I know that we talk a lot about support for Java and Node, but as the other languages are, maybe the others are also inconsistent with the behavior that Peter considers expected. Is there a way for the proxy to fully accomplish the task?
@pvlugter  I don't like to keep just context.fail () semantically for me it's like I'm returning an HTTP 500 to the client, when in fact my intention would be an HTTP 400 Bad Request

marcellanz: I think it's not. I don't say to do it that way but that it's a possibility and could be considered.
The sequence of how entities get their state is well defined, sequential and it would not be different from any snapshot or event the entity would apply instructed by the proxy right before it gets any new events or a command from the proxy. Also, and especially with the event sourcing procotol, the proxy does not know what is inside a snapshot or an event, it could be anything. The entity in this regard has to manage its state always in a consistent manner. So just food for thoughts.

I agree not to implement anything complicated, the simpler the better.

marcellanz: @sleipnir the current Go support is implemented by the spec and docs and has the described flaw right now.

sleipnir: Hmm ... I don't know, for me the premise breaks and the premise breaks so to implement this we must change the premise, "proxy manages the state of the application", "we move to the proxy shares the state with the application", I definitely don't know, I would have to think about it a lot to have a plausible answer and I have a lot of headaches at the moment. :(

marcellanz: @sleipnir I hope not because of the idea shared by me before. I'm pretty sure we might not are aligned or this is a misunderstanding. What I wrote about using snapshots to revert on a `context.fail` is an optimization of what Peter has described with:

>The usual approach is to wait for events to be persisted successfully before updating the in-memory state, but here **that would require additional communication from the proxy**, but the entity could still have atomicity for handlers by applying after command processing.

where events would be applied after the command and its events got back to the proxy and the proxy would send those events to the user entity to get applied, or I think, the proxy would re-establish the state of the entity after a `context.fail` ed entity error.

I'm curious if we have multiple talking points here, one of
 **a)** what semantics does `context.fail` have regarding the local state, and
 **b)** how should the emitted events get applied, during or after the command succeeded?

I think for **b)** we have a relevant aspect of what is described why to have the local state changed during the time the command handler is running. One is to validate the event can be applied and the other is to "_updates the state so that subsequent processing in the command handler can use it_" defined in the documentation:

>Invoking emit will immediately invoke the associated event handler for that event - this **both validates that the event can be applied to the current state, as well as updates the state so that subsequent processing in the command handler can use it**.

If events are not applied to local state, it could be difficult to a command handler go get the right values of the local state after the `context.emit` applied the to the local state.

Perhaps we could discuss these issues and ideas on the next contributor call :)

sleipnir: Yes @marcellanz , it would be nice to discuss this in a collaborator call, if you participate in any please update the minutes of the meeting, I can rarely participate but I like to read about what was discussed at the meeting.

pvlugter: Updated java-support to only apply events after successful command processing. This still validates and applies the emitted events; if any of them fail then it will crash the entity before the reply is sent.

I think it works to have `context.fail` mean fail the current command (without applying or persisting any events). This should only be available to command contexts.

The protocol does allow returning an action failure along with events, snapshots, or effects — which will be applied by the proxy. If this should never be the case for language supports, then we may want to adjust the protocol so this is not supported.

marcellanz: Thanks. I'll implement the same for Go Support.

marcellanz: @sleipnir we might have to check to have the same semantics with other language support libraries.

ralphlaude: @pvlugter thanks for adapting the java support. i will updated for other event-sourced based model like CRUD.

sleipnir: Thanks @pvlugter 
@marcellanz @pvlugter I will check out Kotlin, Dart, C#, Springboot I just need to update the version of the java-support library, so it would be interesting to release a release of java-support with this fix (alias it would be nice to release a release with everything that has been done since the last version ), perhaps this is the same case for Kotlin. I'm going to take a look at Python too, but if you want to see that for me that's fine. I think that would be it for now, if anyone remembers somewhere else to look let me know.

marcellanz: thanks @sleipnir, I can check the python-support.

ralphlaude: @marcellanz really great for finding the bug!

pvlugter: I can publish another 0.5.x release for java-support with the failure handling changes. I was originally going to wait for the metadata and projection changes to be finished before next releases, but no problem to have extra releases.

marcellanz: @sleipnir The Python Support does apply events after the command handler ran, `context.emit` just collects the event:
https://github.com/cloudstateio/python-support/blob/v0.1.1/cloudstate/evensourced_servicer.py#L100-L103

no change is needed here IMO.

sleipnir: Very good

pvlugter: java-support 0.5.2 is published to bintray and synced to maven central

marcellanz: @pvlugter I tried to run the TCK matrix and the docker release images where not available for 0.5.2. I think the release travis build [did not run through](https://travis-ci.com/github/cloudstateio/cloudstate/jobs/368079957#L496).

pvlugter: I've only released java-support 0.5.2. The versions are similar, but proxy (and TCK), java-support, and node-support are on separate release cycles. The `latest` version of the docker images will have the changes. I didn't release cloudstate core, as the metadata and projection work is in progress and eventing has been disabled as part of this. Eventing may not be used anywhere, so can certainly release if it's needed. 

On the travis build failure: since it's the first time using the new publish via bintray, the maven central sync still needed to be set up — it can't be enabled until something is published first. Bintray have approved the sync now, and subsequent releases will work fine.

PR #394, opened 2020-07-21
add missing conf so akka can find tables

coreyauger:
We need to let akka know which tables to use.

### DIFF
--- a/proxy/spanner/src/main/resources/application.conf
+++ b/proxy/spanner/src/main/resources/application.conf
@@ -10,6 +10,10 @@ akka {
   }
 
   persistence {
+    journal {
+      plugin = "akka.persistence.spanner.journal"
+    }
+
     snapshot-store {
       plugin = "akka.persistence.spanner.snapshot"
     }
--- a/proxy/spanner/src/main/resources/application.conf
+++ b/proxy/spanner/src/main/resources/application.conf
@@ -19,27 +19,19 @@ akka {
     }
 
     spanner {
-      journal {
-        # name of the table to use for events
-        table = "journal"
-        table = "journal_"${?SPANNER_TABLE_PREFIX}
-
-        # name of the (interleaved) table used for tags
-        event-tag-table = "tags"
-        event-tag-table = "tags_"${?SPANNER_TABLE_PREFIX}
-
-        # name of the auxilary table used to store information about deletes
-        deletions-table = "deletions"
-        deletions-table = "deletions_"${?SPANNER_TABLE_PREFIX}
-      }
-
-      snapshot {
-        table = "snapshots"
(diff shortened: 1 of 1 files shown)


coreyauger:
Thanks @pvlugter that makes more sense.

coreyauger:
Think this is what we are looking for now.

coreyauger:
doh! .. ok hows this?

pvlugter:
I've made some changes in #396 so that the tests pass. I've rebased this PR.

proxy/spanner/src/main/resources/application.conf
```diff
@@ -10,15 +10,29 @@ akka {
   }
 
   persistence {
-    journal {
-      plugin = "akka.persistence.spanner.journal"
-    }
-
```
pvlugter: I think we still need this plugin setting. I'll add it back.

proxy/spanner/src/main/resources/application.conf
```diff
@@ -19,6 +19,24 @@ akka {
...
+
+        # name of the auxilary table used to store information about deletes
+        deletions-table = "deletions"
+        deletions-table = "deletions_"${?SPANNER_TABLE_PREFIX}
+      }
+
+      snapshot {
+        table = "snapshots"
+        table = "snapshots_"${?SPANNER_TABLE_PREFIX}
+      }
```
pvlugter, line 39: All these table names are also defined further down the file under `cloudstate.proxy.spanner`. Maybe we should have the table names only defined once, and use setting references?

coreyauger: That makes sense to me.

pvlugter: Also, these tables will be called `journal_`, `tags_`, and so on, when `SPANNER_TABLE_PREFIX` is not set. The initial setting will always be overridden given the string prefix (different from if it was just `table = ${?TABLE}`. I don't know if that matters (maybe table names can't end with an underscore?) if the prefix var is always set — although could then be better to make it required.

coreyauger: Lets loop @hseeberger in here... I believe these are his names.

pvlugter: Okay. Have updated to reference the `cloudstate.proxy.spanner` settings, like for project, instance, database. Removed the redundant settings.

hseeberger:
@coreyauger and @pvlugter, nice job!

proxy/spanner/src/main/resources/application.conf
hseeberger: I'm not sure about the potentially trailing underscore. Let's get this merged quickly to fix the issue at hand and then deal with that underscore later.

PR #408, opened 2020-08-12
Extend the minikube smoke tests

### LINKED ISSUE #379: Add in kubernetes test to native image build
Right now, we only test the native image build against the TCK, we don't test actually running it in Kubernetes. This means a bunch of things don't get exercised:

* Kubernetes cluster bootstrap
* Remoting
* Serialization of remotely sent messages

I'm 110% positive that the current native image build therefore won't work in Kubernetes.

We already have the minikube smoke tests, we should use that, but we should do a little more than what they currently do:

* Scale up to multiple nodes.
* Run a few more tests to ensure serialization of all messages sent over remoting works.

pvlugter:
Resolves #379. Native image builds fixed in #405.

Extend the minikube smoke tests: include tests for postgres and cassandra stores, scale service and test multiple entities for checking clustered nodes. Build non-native in Travis. Add native-image tests to CircleCI.


pvlugter:
Extended smoke tests should be stable enough now. Tests cover in-memory, postgres, and cassandra stores, native image testing in CircleCI, and scaled to 2 nodes to test akka cluster and remoting (Travis would sometimes fail when 3 nodes).

pvlugter:
Merging so this can be tested on other PRs too. And we can see how stable the tests are.

jroper:
Nice!

bin/run-java-shopping-cart-test.sh
```diff
@@ -0,0 +1,250 @@
...
+}
+
+# Wait for the deployment to be available
+echo
+echo "Waiting for deployment to be ready..."
+kubectl wait --for=condition=available --timeout=1m deployment/$deployment || fail_with_details
+kubectl get deployment $deployment
+
+# Scale up the deployment, to test with akka clustering
+# Travis can fail with 3 nodes, only scale to 2
```
jroper, line 182: Perhaps we should remove the smoke tests from the travis build?

pvlugter: Yeah, we could run all of these on CircleCI instead. Looks like we've already extended this month's build credits, so we may need to trim back the testing. We could see how it goes.

pvlugter: Updated in #410. Tried the default machine size first, but that can't scale to 3 nodes, like Travis. The large machine (which we pay for) works fine.

PR #410, opened 2020-08-18
Smoke tests in CircleCI

pvlugter:
Trying the non-native smoke tests in CircleCI, scaled to 3 nodes. Currently on the default machine size, to see if that works.


pvlugter:
Default machine doesn't work (can't scale to 3 nodes, like Travis) but the large machine works fine.

PR #414, opened 2020-08-26
Remove obsolete scala-support

pvlugter:
Currently an outdated version of the java-support.

New scala-support will developed at https://github.com/cloudstateio/scala-support and build on common JVM support.


viktorklang:
Makes sense @pvlugter—LGTM!

PR #416, opened 2020-08-26
More asciidoc conversions

pvlugter:
More asciidoc conversions and fixes:

- all snippets converted to includes (based on #415)
- convert and fix some more xrefs
- more paradox/markdown to asciidoc conversions

### DIFF
--- a/docs-asciidoc/shared-content/modules/ROOT/pages/architecture.adoc
+++ b/docs-asciidoc/shared-content/modules/ROOT/pages/architecture.adoc
@@ -17,7 +17,7 @@ The gRPC protocol spoken between the Akka sidecar and the user code is a Common 
 * The second part of the IR is **a pluggable entity type protocol:** Each entity type defines its own gRPC protocol for communicating between the sidecar and the user function. Here is a snippet of the event sourcing protocol:
 +
 [source,proto]
-....
+----
 service EventSourced {
     rpc handle(stream EventSourcedStreamIn) returns (stream EventSourcedStreamOut) {}
 }
@@ -43,7 +43,7 @@ message EventSourcedReply {
     }
     repeated google.protobuf.Any events = 3;
 }
-....
+----
 
 When a command for an entity arrives, the following messages are sent using this protocol:
 
@@ -58,14 +58,14 @@ The user function is expected to hold the current state of the entity in the con
 The `Command` message contains, among other things, the name of a gRPC rpc call that is being invoked - this RPC call was declared during the discovery phase. It also contains the payload of that gRPC call, along with an extracted entity id, that identifies which entity the call is for. Entity ids are declared through the use of a Protobuf field extension, here is an example user function message that declares an entity id:
 
 [source,proto]
-....
+----
 message AddLineItem {
     string user_id = 1 [(.cloudstate.entity_key) = true];
     string product_id = 2;
     string name = 3;
     int32 quantity = 4;
 }
-....
+----
 
 === Support for HTTP and JSON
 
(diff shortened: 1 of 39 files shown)


pvlugter:
Hmm, the docs build failure is interesting. I thought the paradox snippet labels and asciidoc include tags could co-exist. They can currently, but only because of a bug in the regex in paradox for after the label, which currently allows non-`w` characters, rather than what was probably intended as non-word characters. Tags with `w`s are currently failing. Both the asciidoc tags and the paradox labels are required to be at the end of the line, so we can't reverse the order. We could update paradox to allow anything after the label (after whitespace), to make this work. Or a hack for now, since we won't want both docs to co-exist for long, is to not use labels/tags with `w`s

pvlugter:
@rstento, I don't know if we'll be doing some more paradox to antora conversions, but some observations: there's been a partial conversion, along with moving docs to common shared includes, including extra content from the readme, and changing some of the doc content. More manageable and less error-prone would be do a complete one-to-one conversion of the docs, which could all be automated (including paradox directives), and then start making the changes and additions.

pvlugter:
> we'll obviously have to do an editing pass after this is merged. Pretty easy to miss minor formatting errors or typos

@rstento, I'm hoping that antora will warn or fail if xrefs or includes are incorrect. Along those lines, and I'll look up possibilities, but you may know already:

- How do we enable xref validation? Paradox checks all refs, including fragments to section anchors within pages, but it looks like asciidoc is allowing incorrect xrefs, as I ended up correcting some that I noticed. Same for external link validation.

- Can leading whitespace be trimmed from code snippet includes? Currently any indentation is included. Would be cleaner to have the unnecessary leading whitespace removed, like paradox does.

- If the javascript example tests are built, then antora has much longer build times, I assume because it indexes everything under `node_modules`. Can we exclude directories like `node_modules`, even when they're under `examples`? We could have just the source files under antora, linked to a build somewhere else, if we can't do anything about this.

pvlugter:
Have worked around the snippet label/tags — just uppercased the tags with lowercase `w`. This is still taking advantage of the bug in paradox that allows the asciidoc tags to be at the end of the paradox snippet label lines, which we can leverage for now while both docs are around.

pvlugter:
Answering some of the questions:

- Leading indentation can be stripped with the `indent=0` option. Added this to all source includes in #417

- For xref validation there's https://gitlab.com/antora/xref-validator. Not sure if it checks fragments. Need to try it out. Also:
  > This generator is a short-term solution to provide xref validation for Antora. It will eventually be replaced or rewritten to perform validation that’s more thorough and better integrated into the site generation process.

- For link validation there's https://github.com/gaurav-nelson/asciidoc-link-check

- For ignoring `node_modules` in the js example, to avoid long build times, antora will ignore dot files. Can get npm to install to a different prefix, but doesn't seem straightforward to make that a default, as compared with other tools like yarn. But having the dependencies in a hidden folder by default could be useful. On my machine the antora build is over 30 seconds longer with all the extra files.

rstento:
@Peter Vlugter <peter.vlugter@lightbend.com>, I've not yet found a
converter that does even 70% of the conversion correctly. Probably because
we're using a flavor of markdown with some special markup for paradox.
However, I've had no problem doing the conversion partly automated and then
combining through manually until I got to the Cloudstate doc, which had
complex constructions that simply couldn't be fixed with search and replace.

Ruth Stento
User Experience Lead
Lightbend

512-696-1448
<http://lightbend.com>

rstento:
@Peter Vlugter <peter.vlugter@lightbend.com>, we're using the
asciidoc-link-check in other projects. It takes some tweaking when it flags
links as broken and they aren't. But in general works fine.

Thanks for finding and fixing these other misc. issues!

Ruth Stento
User Experience Lead
Lightbend

512-696-1448
<http://lightbend.com>

pvlugter:
> I've not yet found a converter that does even 70% of the conversion correctly. Probably because we're using a flavor of markdown with some special markup for paradox. However, I've had no problem doing the conversion partly automated and then combining through manually until I got to the Cloudstate doc, which had complex constructions that simply couldn't be fixed with search and replace.

Yes, the paradox directives need to be converted separately, and it could be easier to do these first. There is an obvious pattern to the conversions though, and I used find and replace with regular expressions for these changes.

I expect [pandoc](https://pandoc.org/) is still one of the best tools for doc conversions, and is extensible.

Issue #359, opened 2020-06-15
TCK – verify entity passivation


marcellanz:
An [issue](https://github.com/cloudstateio/go-support/issues/27) reported in v0.1.0 of the Go Support brought up a testcase the TCK could verify for state models. 

The proxy sets the timeout for entity passivation to 30s for now. The property [`cloudstate.proxy.passivation-timeout`](https://github.com/cloudstateio/cloudstate/blob/v0.5.1/proxy/core/src/main/resources/reference.conf#L17) can be set to a different value for the TCK running the proxy.

**Tasks**
- [x] verify entity passivation for eventsourcing with the TCK
- [x] verify entity passivation for CRDTs. I will include this for CRDTs with the new CRDT TCK [this issue](https://github.com/cloudstateio/go-support/issues/24) to be solved.

[Go issue#27](https://github.com/cloudstateio/go-support/issues/27#issuecomment-643663849) demonstrates how it can be verified.

marcellanz:
On passivation this message is logged by the proxy:
```
2020-06-14 23:42:13.486 INFO akka.actor.LocalActorRef - Message [akka.actor.Status$Failure] to Actor[akka://cloudstate-proxy/system/sharding/ShoppingCart/50/mrcl#-1089585546] was not delivered. [1] dead letters encountered. If this is not an expected behavior then Actor[akka://cloudstate-proxy/system/sharding/ShoppingCart/50/mrcl#-1089585546] may have terminated unexpectedly. This logging can be turned off or adjusted with configuration settings 'akka.log-dead-letters' and 'akka.log-dead-letters-during-shutdown'.
```

What does it mean by the proxies perspective?

pvlugter:
Yes, I've also noticed the deadletter on passivation and have been meaning to look at it.

viktorklang:
Good, thanks @pvlugter!

Thanks for opening this issue, @marcellanz

pvlugter:
I've moved the dead letter log to issue #389

pvlugter:
Entity passivation in the language supports is now verified through the new TCK tests which call the subprotocols directly. We can also test the passivation timeouts from the language supports when based on #486. I'll close this issue as resolved.

PR #419, opened 2020-08-28
Extend TCK with event-sourced model test

pvlugter:
First step in evolving the TCK, #316.

Moves the testing infrastructure for the TCK to the common testkit, and updates it to use a send/expect style (and closer to the generic testkit that we've discussed).

Adds a model test for event-sourced entities. This model test requires implementing simple event-sourced entities, where the tests send actions to take (emitting events, forwarding, side effects, failing). The TCK determines which tests to run based on which entities/services have been implemented — marking unimplemented tests as pending.
 
The event-sourced model test has only been implemented for Java support so far. This uncovered one small bug, where side effects were not actually being recorded. I want to look at fixing https://github.com/cloudstateio/cloudstate/issues/375#issuecomment-672336020 and updating the TCK tests to reflect these changes — currently have FIXMEs where the updated state should be returned — before implementing the new tests for other language supports. And then add a state model test for CRDT entities as well.

Event-sourced tests also cover verifying passivation on the language support side (#359) by testing the event-sourced entity protocol directly.

### DIFF
--- a/protocols/tck/cloudstate/tck/model/eventsourced.proto
+++ b/protocols/tck/cloudstate/tck/model/eventsourced.proto
@@ -1,0 +1,123 @@
+// Copyright 2019 Lightbend Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//
+// == Cloudstate TCK model test for event-sourced entities ==
+//
+
+syntax = "proto3";
+
+package cloudstate.tck.model;
+
+import "cloudstate/entity_key.proto";
+
+option java_package = "io.cloudstate.tck.model";
+
+//
+// The `EventSourcedTckModel` service should be implemented in the following ways:
+//
+// - The entity persistence-id must be `event-sourced-tck-model`.
+// - Snapshots must be configured for every 5 events.
+// - The state of the entity is simply a string.
+// - Event and snapshot string values are wrapped in `Persisted` messages.
+// - The snapshot handler must set the state to the value of a `Persisted` message.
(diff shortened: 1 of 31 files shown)


tck/src/it/resources/application.conf
```diff
@@ -25,7 +25,7 @@ cloudstate-tck.combinations = [{
     }
   }
 
-  frontend {
+  service {
```
marcellanz, line 28: nice to have that renamed to service.

tck/src/main/resources/reference.conf
```diff
@@ -11,7 +11,7 @@ cloudstate.tck {
     port = ${?TCK_PROXY_PORT}
   }
 
-  frontend {
+  service {
     hostname = "127.0.0.1"
     hostname = ${?TCK_FRONTEND_HOST}
```
marcellanz, line 16: should `TCK_FRONTEND_HOST` be renamed too to `TCK_SERVICE_HOST `?

protocols/tck/cloudstate/tck/model/eventsourced.proto
```diff
@@ -0,0 +1,123 @@
...
+// Each `RequestAction` is one of:
+//
+// - Emit: emit an event, with a given value.
+// - Forward: forward to another service, in place of replying with a Response.
+// - Effect: add a side effect to another service to the reply.
+// - Fail: fail the current `Process` command.
+//
+message RequestAction {
+  oneof action {
+    Emit emit = 1;
```
marcellanz, line 72: Nice. I've chosen separate service methods for synthetic (model) [CRDT tests](https://github.com/cloudstateio/go-support/blob/feature/crdt_support/protobuf/tck/tck_crdt.proto#L30) for which `RequestAction`s are used here. As it increases the TCK model implementation for certain user languages and also you've chosen a similar approach here, I'll change that to a "process" service method per CRDT type too.

tck/src/main/resources/reference.conf
pvlugter: Yes, forgot about this. I was going to add `TCK_SERVICE_HOST` as well as keep the old `TCK_FRONTEND_HOST` in case it's being used. But I don't see anything that uses these env vars under cloudstate repos, so let's just rename.

protocols/tck/cloudstate/tck/model/eventsourced.proto
pvlugter: Yes, I thought this would be the easiest to implement, and especially since there are tests that want to have multiple actions. And hopefully also easier to change the tests without needing to change the implementation.

PR #420, opened 2020-08-28
Asciidoc: add xref and link validation

pvlugter:
- Add xref validation with `@antora/xref-validator` (checks on every build).
- Add external link validation with `asciidoc-link-check` (run manually to check).

The xref validator did uncover some more incorrect xrefs, and typos.

@rstento, note: the xref validator fails on a bunch of xrefs from the shared content to the main docs content. I've added the `docs` component into these xrefs so that they're considered valid, but it highlights that the shared content has links to other content that would also need to be included anywhere that these pages are used. See the ones with `xref:docs:`.

Also fix some remaining markdown-style links.


docs-asciidoc/shared-content/modules/ROOT/pages/working.adoc
```diff
@@ -1,4 +1,4 @@
-Cloudstate provides a good developer experience by implementing the low level details that allow you to focus on your business logic. When using Cloudstate open source, you need to provide your own Kubernetes environment and provision Cloudstate to run in it. Lightbend also provides a https://lbcs.io[paid Cloudstate service] that manages the Kubernetes environment for you. 
+Cloudstate provides a good developer experience by implementing the low level details that allow you to focus on your business logic. When using Cloudstate open source, you need to provide your own Kubernetes environment and provision Cloudstate to run in it. Lightbend also provides a https://lbcs.io[paid Cloudstate service] that manages the Kubernetes environment for you.
```
marcellanz, line 1: how differs lbcs.io from console.cloudstate.com? Are those instances the same? I meant the latter is the "production" one but I'm not sure.

rstento:
@marcellanz, good catch! These links should be to console.cloudstate.com.
This was written before that site was live. Also, maybe maybe we should
change "paid Cloudstate service" to just "Cloudstate service", the name
will be changing in the future as well.

Ruth Stento
User Experience Lead
Lightbend

512-696-1448
<http://lightbend.com>

docs-asciidoc/shared-content/modules/ROOT/pages/working.adoc
pvlugter: > @marcellanz, good catch! These links should be to console.cloudstate.com.
This was written before that site was live. Also, maybe maybe we should
change "paid Cloudstate service" to just "Cloudstate service", the name
will be changing in the future as well.

Rather than console.cloudstate.com (which will just be the login page), how about we use https://www.lightbend.com/cloudstate-by-lightbend? (Actually, that already redirects to https://www.lightbend.com/akka-serverless)

pvlugter: I've updated to the informational link on the Lightbend site, which is now the Akka Serverless page. I've left the link name as "Cloudstate service".

PR #421, opened 2020-09-02
Antora

coreyauger:
Fix the `make build` so that the required jsdocs work for everyone.

### DIFF
--- a/docs-asciidoc/Makefile
+++ b/docs-asciidoc/Makefile
@@ -37,6 +37,8 @@ html-author-mode:
 apidocs:
 	cd ${root_dir} && sbt java-support/doc
 	cp -r ${root_dir}/java-support/target/api ${site_dir}/docs/current/java/api
+	cd ${root_dir}/node-support && npm i
+	npm i -g jsdoc
 	cd ${root_dir}/node-support && npm run jsdoc
 	cp -r ${root_dir}/node-support/apidocs ${site_dir}/docs/current/javascript/api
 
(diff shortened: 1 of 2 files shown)


docs-asciidoc/Makefile
```diff
@@ -35,6 +35,8 @@ html-author-mode:
 apidocs:
 	cd ${root_dir} && sbt java-support/doc
 	cp -r ${root_dir}/java-support/target/api ${site_dir}/docs/current/java/api
+	cd ${root_dir}/node-support && npm i
+	npm i -g jsdoc
```
pvlugter, line 39: Does it need to be installed globally as well?

docs-asciidoc/Makefile
```diff
@@ -35,6 +35,8 @@ html-author-mode:
 apidocs:
 	cd ${root_dir} && sbt java-support/doc
 	cp -r ${root_dir}/java-support/target/api ${site_dir}/docs/current/java/api
+	cd ${root_dir}/node-support && npm i
```
pvlugter, line 38: Sure, let's add the install. The CI scripts currently do this in a before script, so the old docs setup requires dependencies to already have been installed, which is why not added here originally.

pvlugter (commented):
Could we just have this based on `antora` branch, without the merged from master.

pvlugter:
I've rebased the `antora` branches

docs-asciidoc/Makefile
coreyauger: yea probably not necessary.

Issue #335, opened 2020-05-24
Protoc plugin to generate stub for specified language from gRPC descriptor


michaelpnash:
A suggestion for a plugin for protoc  compiler which generates a stub for a given language support from a gRPC descriptor

marcellanz:
This would be helpful also for languages where the generated protobuf artefacts are not fully complete; like the one for Go where service descriptors are not exported to user code. In this case, service names and methods are not available in a typesafe manner. Even the new protocolbuffer support for Go does not support that: [ServiceName and Method names are private](https://github.com/cloudstateio/go-support/blob/feature/crdt_support/tck/proto/crdt/tck_crdt.pb.go#L3487).

sleipnir:
How would it be? I believe that there should be several plugins, one for each support language and maybe this is a hard job. It would be welcome of course, but I think a simpler way is to get tools like Cloudstate Community CLI or lbcs CLI to do that. Because I think that in this case it would just be compiling the proto files for each language with some code generators on top.
Just thinking....

marcellanz:
I think what his issue describes is a generic plugin to scaffold anything between a gRPC protobuf file and what the user has to implement just _before_ he can start to write his business code. Even if hard to implement, it has to be done only once.
The Cloudstate Community CLI would benefit of that too I think.

sleipnir:
I am thinking about the implementation and I don't see how a generic plugin could be in the protoc. But any wide entryway helps

PR #424, opened 2020-09-04
Scala Language Support – Counter Sample Straw Man [NO MERGE]

ryanhanks:
First draft on some rough ideas around how scala language support could look. This only intended to serve as a discussion piece.

This example takes on a certain implementation style where behavior is organized by state. We can look at other styles in another discussion PR.

Opening this up early for comments, I'll try to flesh things out as questions and comments come in, and can continue to elaborate if there's something of interest here.

### DIFF
--- a/protocols/frontend/cloudstate/entity_key.proto
+++ b/protocols/frontend/cloudstate/entity_key.proto
@@ -28,3 +28,18 @@ option go_package = "github.com/cloudstateio/go-support/cloudstate/;cloudstate";
 extend google.protobuf.FieldOptions {
     bool entity_key = 50002;
 }
+extend google.protobuf.MessageOptions {
+    EntityMessageOptions entity = 50002;
+}
+
+extend google.protobuf.OneofOptions {
+    bool api_response_value = 50002;
+}
+
+message EntityMessageOptions {
+    bool api_request = 50002;
+    bool api_value_response = 50003;
+    bool api_value = 50004;
+    bool domain_event = 50005;
+    bool domain_state = 50006;
+}
--- a/samples/scala-counter/src/main/protobuf/counter/counterapi.proto
+++ b/samples/scala-counter/src/main/protobuf/counter/counterapi.proto
@@ -1,0 +1,133 @@
+syntax = "proto3";
+import "cloudstate/entity_key.proto";
+
+package com.example.counter.api;
+
+service CounterService {
+
+  rpc Initialize(InitializeRequest) returns (InitializeResponse) {}
+
+  rpc Start(StartRequest) returns (StartResponse) {}
+
+  rpc Increment(IncrementRequest) returns (IncrementResponse) {}
+
+  rpc Stop(StopRequest) returns (StopResponse) {}
+
+  rpc GetCounter(GetCounterRequest) returns (GetCounterResponse) {}
+
+}
+
+/*
+ * Start
+ */
+
+message InitializeRequest {
+  option (cloudstate.entity).api_request = true;
+  string counter_id = 1 [(.cloudstate.entity_key) = true] ;
+  uint32 initial_value = 2;
+}
+
+message InitializeResponse {
+  option (cloudstate.entity).api_value_response = true;
+  oneof counter  {
+    UninitializedCounter uninitializedCounter = 1;
+    ActiveCounter activeCounter = 2;
+    StoppedCounter stoppedCounter = 3;
+  }
+}
+
+/*
+ * Start
+ */
+
+message StartRequest {
+  option (cloudstate.entity).api_request = true;
+  string counter_id = 1 [(.cloudstate.entity_key) = true];
+  uint32 max_value = 2;
+}
+
+message StartResponse {
+  option (cloudstate.entity).api_value_response = true;
+  oneof value  {
+    UninitializedCounter uninitializedCounter = 1;
+    ActiveCounter activeCounter = 2;
+    StoppedCounter stoppedCounter = 3;
+  }
+}
+
+
+
+message IncrementRequest {
+  option (cloudstate.entity).api_request = true;
+  string counter_id = 1 [(.cloudstate.entity_key) = true];
+  uint32 amount = 2;
+}
+
+message IncrementResponse {
+  option (cloudstate.entity).api_value_response = true;
+  oneof value {
+    UninitializedCounter uninitializedCounter = 1;
+    ActiveCounter activeCounter = 2;
+    StoppedCounter stoppedCounter = 3;
+  }
+}
+
+/*
+ * Stop
+ */
+
+message StopRequest {
+  option (cloudstate.entity).api_request = true;
+  string counter_id = 2;
+}
+
+message StopResponse {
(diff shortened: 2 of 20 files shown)


pvlugter (commented):
@ryanhanks, thanks for putting this together! Great to some different ideas being explored.

Interesting to explore specifying extra options in the protocol with more code automatically generated, to have type safe APIs to implement. Also relates to this issue for generating stubs: https://github.com/cloudstateio/cloudstate/issues/335.

Having the built-in pattern for resource-style responses, automatically returning the updated state, is an interesting simplification. Do you think we would still need a way to specify other responses as well? (could be something like `CommandEffects.accept(event).thenReply(newState => response)`)

And organising behaviour by different states is also interesting, having something that moves towards encoding state machines directly for the protocol.

ralphlaude:
@ryanhanks great for exploring ideas regarding the code generation with already built-in patterns. i like the way the state is always passed in the handler methods.

samples/scala-counter/src/main/protobuf/counter/counterapi.proto
```diff
@@ -0,0 +1,135 @@
+syntax = "proto3";
+import "cloudstate/entity_key.proto";
+
+package com.example.counter.api;
+
+service CounterService {
+
+  rpc Initialize(InitializeRequest) returns (InitializeResponse) {}
```
viktorklang, line 8: @ryanhanks Can't the api_request and api_response be inferred from parameter types and return values?

protocols/frontend/cloudstate/entity_key.proto
```diff
@@ -28,3 +28,18 @@ option go_package = "github.com/cloudstateio/go-support/cloudstate/;cloudstate";
...
 }
+extend google.protobuf.MessageOptions {
+    EntityMessageOptions entity = 50002;
+}
+
+extend google.protobuf.OneofOptions {
+    bool api_response_value = 50002;
+}
+
+message EntityMessageOptions {
```
viktorklang, line 39: Aren't these all these values mutually exclusive? I.e. a `oneof` or `enum` would be more fitting?

samples/scala-counter/src/main/protobuf/counter/counterapi.proto
ryanhanks: To make sure I understand correctly, are you suggesting these rpc method definitions are redundant, given that these types are now specified by the handler implementation? I was envisioning the trait that the handler impl inherits from to be driven from the service definition, so specifying the types here is necessary in this approach in order to generate the underlying trait.

Wondering if I'm misunderstanding your question though..

protocols/frontend/cloudstate/entity_key.proto
ryanhanks: That's correct, thanks for noticing. I threw these bools in here a bit haphazardly, not knowing whether or not the idea of introducing new / additional metadata through extensions is something we're open to... Happy to iterate on / refine / further explore this if there's interest in this direction 👍

ryanhanks: It could spin-off into it's own discussion, but I wouldn't want to have movement on the Scala language support be entirely hinged on that if that discussion is projected to take a long time. One idea to consider is that we take a first pass at implementing Scala support with the cloudstate API defined exactly as it currently is, while at the same time trying to move forward the discussion around new metadata. In coming up with my ideal example the ideas were hard to separate, and I do find them to be very closely related, so my preference would be to have a first-pass that includes the introduction of some new metadata, as long as that doesn't have the potential to draw out into a super long process.

ryanhanks:
> Having the built-in pattern for resource-style responses, automatically returning the updated state, is an interesting simplification. Do you think we would still need a way to specify other responses as well? (could be something like CommandEffects.accept(event).thenReply(newState => response))

How do you mean by specify other responses? The example here isn't intended to convey that mapping might be entirely automated, you'll notice that each state handler implements a mapper object that's assumed to be used in the default execution by the underlying implementation:

```
  /*
  * Response object mapper
  * */

  override def mapDomainCounterToApiCounter(activeCounter: ActiveCounter): CounterApi.Counter =
    CounterApi.Counter.activeCounter(activeCounter.value, activeCounter.maxValue)
```

The assumption here is that each state handler can provide a default mapping from one of the internal states to some external response, but that each state handler is responsible for providing a reply if it is the resultant state. 

If further customization is desired, one of the underlying buildResponse methods could be overridden (these would have different default behavior that calls the mapper method rather than throwing, just copying and pasting from the example as it exists now):

```
  /*
  * Response builders
  * */

  def buildInitializeResponse(activeCounter: ActiveCounter, initializeRequest: InitializeRequest): InitializeResponse = throw new UnsupportedOperationException("No response")

  def buildStartResponse(activeCounter: ActiveCounter, startRequest: StartRequest): StartResponse = throw new UnsupportedOperationException("No response")

  def buildIncrementResponse(activeCounter: ActiveCounter, incrementRequest: IncrementRequest): IncrementResponse = throw new UnsupportedOperationException("No response")

  def buildStopResponse(activeCounter: ActiveCounter, stopRequest: StopRequest): StopResponse = throw new UnsupportedOperationException("No response")

  def buildGetCounterResponse(activeCounter: ActiveCounter, getCounterRequest: GetCounterRequest): GetCounterResponse = throw new UnsupportedOperationException("No response")

```

samples/scala-counter/src/main/protobuf/counter/counterapi.proto
viktorklang: @ryanhanks No, rather the other way around—using the `rpc` definitions you can know what messages are api_requests and which ones are api_responses. I.e. the annotations might be redundant?

protocols/frontend/cloudstate/entity_key.proto
viktorklang: @ryanhanks I think you're right—let's keep the "new possible metadata" as a separate track—makes the Scala support a bit more focused.

samples/scala-counter/src/main/protobuf/counter/counterapi.proto
ryanhanks: Ah, gotcha. Yeah so we could annotate these method descriptors with a method option, and use that if we wanted to make it explicit, and then I think that could eliminate requiring the declarations on the request and response messages 👍

viktorklang: :+1:

pvlugter:
> The assumption here is that each state handler can provide a default mapping from one of the internal states to some external response, but that each state handler is responsible for providing a reply if it is the resultant state.
> 
> If further customization is desired, one of the underlying buildResponse methods could be overridden

Yes, makes sense. That would provide all the customisation needed. And I think returning the updated state in a RESTful resource style is a good default 👍

pvlugter:
@ryanhanks, I'll close this straw man PR now. Ideas for code generation and type-safe APIs are definitely interesting. Any more work on Scala language support can go in https://github.com/cloudstateio/scala-support

PR #425, opened 2020-09-07
Ported operator to Go

jroper:
There's still a bit of work to do, but the main thing I want to see here is if the smoke tests are working.


bin/deploy-cloudstate.sh
```diff
@@ -39,20 +39,20 @@ else
...
+if [ -n "$build_commands" ]
+then
+  sbt "$build_commands"
+fi
+
+kubectl apply --validate=false -f https://github.com/jetstack/cert-manager/releases/download/v0.16.1/cert-manager.yaml
+make -C cloudstate-operator deploy
 echo "Waiting for operator deployment to be ready..."
-if ! kubectl wait --for=condition=available --timeout=2m -n cloudstate deployment/cloudstate-operator
+if ! kubectl wait --for=condition=available --timeout=2m -n cloudstate-operator-system deployment/cloudstate-operator-controller-manager
```
viktorklang, line 52: cloudstate-operator-controller-manager … -provider-factory-bean :)

sleipnir: Why change the namespace from cloudstate to cloudstate-operator-system? Much more verbosity.

jroper: Yeah, `cloudstate-operator-system` doesn't make sense. I'll rename to `cloudstate-system`, the `-system` suffix is a convention for operators that I think is a good idea to follow.

viktorklang:
@jroper Have you been able to get further on the CI side of this?

jroper:
circle ci is now passing.

I haven't updated anything with travis yet, and what that means is:

* Operator tests are not currently being run by travis.
* The manifest generated and uploaded as part of the release is the wrong one.

viktorklang:
@jroper Awesome, so we should be really close to completion?

marcellanz:
> There's still a bit of work to do, but the main thing I want to see here is if the smoke tests are working.

What else is planned for this PR beside "smoke tests are working" and the (circle) CI running?

marcellanz:
>There's still a bit of work to do, but the main thing I want to see here is if the smoke tests are working.

Because of this note, enjoy my comments with caution. I'm pretty sure many points are already known. I collected my comments for the code in this review branch: link_to_branch
as it would be too much work to place them here on every occasion.

Feel free to pick whatever seems to be helpful or makes sense. There are quite a few missed "If with a short statement" statements, where it cleans up boilerplate code. Also, (deeply) nested if/else statements can be rewritten to get the "happy path" in front of the lines and give the reader a flow of what usually should happen, with returns where it doesn't.

Packages should be part of types, so instead of `stores.PostgresStore` one might write `store.Postgres`. Note the singular form of the package name, it's not wrong to have plural package names, but they usually don't have to be.

Function/method argument list length seem to be overly large often in the PR. I think it feels unusual to have so many arguments, or it feels unusual to have them grouped as they are often laid out each on one line. This might be a consequence of long type names or not using the package as part of the type. Long argument lists seem to indicate that the code can be refactored to capture the context of the data passed.

I think many variables in this PR are longer than they needed to be. The code looks like written with a JAVA(?) background carrying with. It's not wrong, just generally unusual from a Go context of view. This applies to many `*.go` files in this PR. Also, as one can read elsewhere, Kubernetes code and style sometimes might differ from what idiomatic Go code looks likes and it might be here, in a k8s operator, normal too.

So `statefulService` migth just be named `service`since we're in a method of the `StatefulServiceReconciler`, what other service would we ever talk about in this context.

Panics: I think panics have to be commented and I've wondered how the correct way to end an operator is, and if it is by panicking.

I have found no bugs, but I also did not review the operators functionality.

Many comments can be read here: https://github.com/golang/go/wiki/CodeReviewComments for self-reviews if needed.

Again, take whatever seems to be helpful as this PR has "still a bit of work to do".

sleipnir:
Beautiful review @marcellanz

sleipnir:
I think it is important that the question of "the best way for the GO language" be verified, Marcel's comments are valuable and should be taken into account. This would demonstrate how well Cloudstate is aligned with everyone's expectations and demonstrate quality in all directions. It would also perhaps interest more Go experts and Go Operators' creators to take a look at Cloudstate

ralphlaude:
@jroper awesome! for migrating the operator to Go. it has surely been a long journey

cloudstate-operator/hack/boilerplate.go.txt
```diff
@@ -0,0 +1,14 @@
+/*
+
```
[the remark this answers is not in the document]
jroper: This file is untouched from what kubebuilder generates when you create a new operator.

jroper:
@marcellanz I think everything you mention is great and we should do it, but it's going to be really difficult to track and reviwe in this PR, so I think the best way forward here is to merge, and then iterate, that will give us a chance to get more specific and flesh out things a bit more in the reviews. We can work concurrently on that.

marcellanz:
> @marcellanz I think everything you mention is great and we should do it, but it's going to be really difficult to track and reviwe in this PR, so I think the best way forward here is to merge, and then iterate, that will give us a chance to get more specific and flesh out things a bit more in the reviews. We can work concurrently on that.

@jroper I agree, great idea. I'll create a PR with the review branch.

PR #434, opened 2020-09-20
Update devcontainer build

pvlugter:
This started failing on master, as the common setup script has been updated. It now has more automatic support for creating the user, so I've switched to this (using the defaults, which are the same as we were providing). But we may want to pin the version of this script, rather than using master.


marcellanz:
tested on GitHub Codespaces validated with:
`cloudstateio/cloudstate-devcontainer@sha256:a8cde3c7f1de7de4fa6231f2e1829523b0918679aec77ccc2687c62fdd220800` running.

Looks good, thanks @pvlugter.

marcellanz (approved):

Issue #413, opened 2020-08-24
Remove ConcurrencyEnforcer


viktorklang:
According to @jroper this is no longer necessary, and for performance reasons should most likely be removed.

pvlugter:
ConcurrencyEnforcer removed in #436

PR #436, opened 2020-09-21
Removes ConcurrencyEnforcer—and by extension also Warmup

viktorklang:
See #413 

Ping @jroper & @pvlugter regarding if we should make an effort to keep Warmup or not, given graalvm native-image


ralphlaude:
@viktorklang nice to remove code! could i know why it is not longer necessary (it is not explained here #413)?

pvlugter (approved):
LGTM

It would be interesting to run load tests with the concurrency enforcer removed. I wonder if it will be easy to end up with some over-parallelisation slowdowns, depending on the concurrency settings in the proxy and user function.

I think it's fine to remove the warmup as it is now. There's the native-images. And for running with hotspot JVMs, using CDS (class data sharing to load the classes from an image) and jaotc (ahead-of-time compile up to tier 2), and then have the JIT compiler continue from there, could be the best way to improve startup.

pvlugter:
One of the go operator tests may be flaky, or just need more time to wait for a check. I'll restart that test.

sleipnir:
> LGTM
> 
> It would be interesting to run load tests with the concurrency enforcer removed. I wonder if it will be easy to end up with some over-parallelisation slowdowns, depending on the concurrency settings in the proxy and user function.
> 
> I think it's fine to remove the warmup as it is now. There's the native-images. And for running with hotspot JVMs, using CDS (class data sharing to load the classes from an image) and jaotc (ahead-of-time compile up to tier 2), and then have the JIT compiler continue from there, could be the best way to improve startup.

@pvlugter  using Openjdk / openJ9 using xshareClasses is a great option, I have used it in production with great results on many of our JVM based systems.

jroper:
Warmup still has advantages, there's a lot of things that get initialised and warmed during warm up beyond JIT, the biggest thing being database connection pools.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
```diff
@@ -253,9 +244,8 @@ final class EventSourcedEntity(configuration: EventSourcedEntity.Configuration,
     currentCommand =
       EventSourcedEntity.OutstandingCommand(idCounter, actorId + ":" + entityId + ":" + idCounter, sender)
     commandStartTime = System.nanoTime()
-    concurrencyEnforcer ! Action(currentCommand.actionId, () => {
-      relay ! EventSourcedStreamIn(EventSourcedStreamIn.Message.Command(command))
-    })
+    relay ! EventSourcedStreamIn(EventSourcedStreamIn.Message.Command(command))
+    statsCollector ! StatsCollector.CommandSent
```
jroper, line 248: I think you can remove stats collector too... it's only used for autoscaling.

pvlugter:
> Warmup still has advantages, there's a lot of things that get initialised and warmed during warm up beyond JIT, the biggest thing being database connection pools.

The warmup looks like it just processes one command. Doesn't seem like that would have any affect on JIT. So it's more about class loading and initialising a database connection as early as possible?

jroper:
Yeah I should have said class loading not JIT. There's probably also a significant amount of other initialisation, eg reflection etc, as various components are loaded and initialised for the first time.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedEntity.scala
viktorklang: @jroper Don't we want autoscaling? :)

viktorklang:
Let's see what happens with the native image bahavior when the warmup is gone. If we need to we'll reintrocue something which achieves the same effect, but extrapolated over more than just the ES code.

PR #439, opened 2020-09-24
Add restart on failure option for event sourced entities

pvlugter:
For #375. This adds support for restarting entities on user failures, for the case of emitted events before fail is called, while allowing events to be applied immediately to have updated state.

Use actor restart to retain the mailbox. The stream relay functionality is moved to a separate actor, so that it can be a stable relay actor ref for the entity when reconnecting the stream. The entity actor still has its own custom command stash. We could use standard actor stash (and persistent actors use a separate stash internally) but because there are only `stash` and `unstashAll` public methods, with several commands queued it would mean un-stashing all of them and then re-stashing messages again, for each command processed. We would also lose the instrumentation for the command stash that we have now. So instead use the (akka private) prepending to the mailbox that actor stash uses, but only unstash back to the mailbox on restarts. Prepending means that we always retain command order, even if there are recently arrived messages in the mailbox.

Update java-support to apply emitted events immediately, reverting back to the earlier behaviour, and use the entity restart option on failures if events have been emitted. Update the TCK model test for event sourced, now that state is updated in responses again. For the emit-then-fail test in the TCK, it currently always expects the restart option to be used. If language supports optimise this by keeping local snapshots, or have a different approach to atomicity, then this test will need to be updated to test either approach. We can wait until we get to that point though.

### DIFF
--- a/java-support/src/main/scala/io/cloudstate/javasupport/impl/Contexts.scala
+++ b/java-support/src/main/scala/io/cloudstate/javasupport/impl/Contexts.scala
@@ -84,9 +84,11 @@ private[impl] trait AbstractClientActionContext extends ClientActionContext {
 
   protected def logError(message: String): Unit = ()
 
-  final def createClientAction(reply: Optional[JavaPbAny], allowNoReply: Boolean): Option[ClientAction] =
+  final def createClientAction(reply: Optional[JavaPbAny],
+                               allowNoReply: Boolean,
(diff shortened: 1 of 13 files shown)


pvlugter:
Note: the CircleCI tests will fail at the moment. There's an issue with payment and so the large machine is not allocated. We'll just need to ignore those tests for now.

jroper (approved):

PR #440, opened 2020-09-24
Remove old kotlin docs link

pvlugter:
Kotlin docs are converted and moved to kotlin-support repo now.


pvlugter:
Note: the CircleCI tests will fail at the moment. There's an issue with payment and so the large machine is not allocated. We'll just need to ignore those tests for now.

PR #446, opened 2020-09-30
Update cloudstate-operator-native manifest

pvlugter:
Looking to get the CI builds working again. I also get different formatting on other automatically generated files for some reason (word wrap differences), but looks like just the the cloudstate-operator-native manifest needs an update for CircleCI builds.

Travis CI builds are also failing, which looks like a corrupt jar file in the cache. I've cleared all the caches, to see if that fixes things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ before_install:
   - sudo apt-get -qq -y install conntrack jq
 env:
   global:
+  - CACHE_NAME=cloudstate-cache-01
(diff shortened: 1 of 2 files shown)


jroper:
That was a result of upgrading controller-gen and I didn't rerun the native manifest generation, we obviously don't have a check to ensure it is run.

pvlugter:
> That was a result of upgrading controller-gen and I didn't rerun the native manifest generation, we obviously don't have a check to ensure it is run.

There is a check, which is why the CircleCI builds have been failing. But not sure why it passed in the PR.

pvlugter:
Clearing the Travis build caches didn't seem to make any difference, with it still downloading old caches — no idea why, caches are shown as empty in the UI and CLI. Trying adding an env var instead, which will force a different cache identifier.

PR #457, opened 2020-10-13
Remove doc content from README

pvlugter:
Nearly all of the content in the README has been moved to the actual docs now (intro and contributing sections).


marcellanz:
I checked the mention of "Inversion of state" to be removed with this PR. But actually it's not mentioned in the README but in [older documentation for 0.5.1](https://github.com/cloudstateio/cloudstateio.github.io/blob/a2e6a4d558854a421f8ac985b6cad92075433e18/old/docs/core/0.5.1/user/features/index.html) not available anymore I think.

Even if not in this PR and regarding loosing anything not transferred to other places, do we keep the term somewhere or has it transformed into a different turn of phrase?

jroper:
Inversion of state management is mentioned here:

https://github.com/cloudstateio/cloudstate/blob/master/docs/src/shared/modules/concepts/pages/index.adoc

pvlugter:
This section: https://cloudstate.io/old/docs/core/current/user/features/index.html#high-level-concepts
Is now here: https://cloudstate.io/docs/concepts/index.html

Doesn't have the `Inversion of state management` title, but all the content is copied over.

jroper (approved):

PR #465, opened 2020-10-16
Start on eventing support

jroper:
This is the old pull request #353, rebased, and got to the point of compilation. Haven't tried running yet. My next major task is to switch to using Akka projections.


proxy/core/src/main/scala/io/cloudstate/proxy/streams/MergeSequence.scala
```diff
@@ -0,0 +1,128 @@
...
+
+package io.cloudstate.proxy.streams
+
+import akka.NotUsed
+import akka.stream.{Attributes, Graph, Inlet, Outlet, UniformFanInShape}
+import akka.stream.stage.{GraphStage, GraphStageLogic, InHandler, OutHandler}
+
+import scala.collection.mutable
+
+object MergeSequence {
```
pvlugter, line 25: Didn't this already get added to akka?

pvlugter:
> My next major task is to switch to using Akka projections.

👍 Sounds good to switch over to Akka projections.

proxy/core/src/main/scala/io/cloudstate/proxy/streams/MergeSequence.scala
jroper: Yes it did, but my initial aim was just to get the branch rebased and compiling, which is what this is.

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
```diff
@@ -193,162 +193,291 @@ class GCPubsubEventingSupport(config: Config, materializer: Materializer) extend
...
-        .map(any => PubsubMessage(data = any.toByteString))
+      Flow[DestinationEvent]
+        .map(transformDestinationEvent)
         .groupedWithin(downstreamBatchSize, downstreamBatchDeadline)
-    } else Flow[ProtobufAny].map(any => PubsubMessage(data = any.toByteString) :: Nil)
+    } else Flow[DestinationEvent].map(event => transformDestinationEvent(event) :: Nil)
 
-  private[this] final def createManualSource(
+  private def sourceToSourceToFlow[In, Out, MOut](f: Source[In, NotUsed] => Source[Out, MOut]): Flow[In, Out, NotUsed] =
+    Flow[In].prefixAndTail(0).flatMapConcat { case (Nil, in) => f(in) }
```
pvlugter, line 217: Seems awkward for akka streams if the best way to connect this is to use empty `prefixAndTail` and substreams. Don't know of a more direct way.

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
```diff
@@ -193,162 +193,291 @@ class GCPubsubEventingSupport(config: Config, materializer: Materializer) extend
...
+
+      }
+      val killSwitch = manageTopicsAndSubscriptions match {
+        case MANUALLY => runManualFlow(subscription, flow)
+        case USING_CRD => futureKillSwitch(createUsingCrdManagedFlow(consumerGroup, subscription, flow))
+        case BY_PROXY => futureKillSwitch(runByProxyManagedFlow(consumerGroup, subscription, flow))
+      }
+
+      new Cancellable {
+        private val running = new AtomicBoolean()
```
pvlugter, line 343: `running` -> `cancelled` here as well?

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/InMemJournal.scala
```diff
@@ -0,0 +1,220 @@
...
+      ms <- messages.get(pid)
+      m <- ms.lastOption
+    } yield m.repr.sequenceNr
+    snro.getOrElse(0L)
+  }
+
+  def eventsByTagQuery(tag: String, fromOffset: Long): Source[EventEnvelope, NotUsed] =
+    // Technically, there's a race condition that could result in messages being dropped here, if a new message is
+    // persisted after this returns, but before the subscriber to the query materializes this stream, that message
+    // will be dropped. But this is only the in memory journal which makes no sense to use in production.
```
pvlugter, line 191: Fair enough that it's likely not needed. If we did want to close this gap, I guess it could retrieve the old messages more as a lazy source once materialized, or something similar.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/InMemJournal.scala
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+/**
+ * INTERNAL API.
+ */
+trait InmemMessages {
+  private var messages = Map.empty[String, Vector[Message]]
+  private var allMessages = Vector.empty[Message]
+  private var offset: Long = 0
```
pvlugter, line 120: Do we need these vars to be thread-safe?

pvlugter (commented):
LGTM. There's a lot there, with some complexity in places. Examples and tests are looking good :+1:

pvlugter:
Failing in the TCK. Unexpected connection type. We could have these go through different probes, so the tests don't affect each other, but it's also useful to know that something wasn't expected.

proxy/core/src/main/scala/io/cloudstate/proxy/EntityDiscoveryManager.scala
```diff
@@ -170,6 +198,21 @@ class EntityDiscoveryManager(config: EntityDiscoveryManager.Configuration)(
...
     }
 
+  private final val elAndPs: Option[(EventingSupport, ProjectionSupport)] = for {
+    projectionSupportClass <- config.eventSourcedSettings.projectionSupport
+  } yield {
+    val projectionSupport: ProjectionSupport = getClass.getClassLoader
+      .loadClass(projectionSupportClass)
+      .asSubclass(classOf[ProjectionSupport])
+      .getDeclaredConstructor(classOf[TypedActorSystem[_]])
+      .newInstance(system.toTyped)
```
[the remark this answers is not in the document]
viktorklang: Yeah, seems like this should use the system.dynamicAccess

jroper:
The TCK failure is because I'd created an event sourced subscriber in the JS shopping cart example.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/InMemJournal.scala
jroper: Yeah, I'll mark them as volatile.

jroper: Actually no, they are only accessed from within an actor.

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
jroper: Related issues:

https://github.com/akka/akka/issues/26592
https://github.com/akka/akka-grpc/issues/550

protocols/example/shoppingcart/products.proto
```diff
@@ -0,0 +1,48 @@
...
+// limitations under the License.
+
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+import "cloudstate/entity_key.proto";
+
+package com.example.shoppingcart;
+
+option go_package = "tck/shoppingcart";
```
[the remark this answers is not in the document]
jroper: I've deleted this file anyway.

proxy/core/src/main/scala/io/cloudstate/proxy/EntityDiscoveryManager.scala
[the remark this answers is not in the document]
jroper: We're not an extension though, so we don't have `ExtendedActorSystem`, we'd have to cast `ActorSystem` to it, is that safe?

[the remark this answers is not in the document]
pvlugter: I think it's always an `ExtendedActorSystem`, so should be safe. Akka codebase has plenty of `.asInstanceOf[ExtendedActorSystem]`. Would make sense to have something better than a cast to access...

proxy/core/src/main/scala/io/cloudstate/proxy/eventing/GooglePubsubEventing.scala
pvlugter: :+1: good to see that there are issues already. Both make sense to have, the utility method and the flow API in akka-grpc. I guess the alternative to the substreams approach is to have a reactive streams publisher/subscriber boundary to connect, although that probably means using the pre-materialize approach as well.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/InMemJournal.scala
pvlugter: Ok, cool.

pvlugter (approved):
LGTM. Nice work!

testkit/src/main/scala/io/cloudstate/testkit/action/InterceptActionService.scala
```diff
@@ -0,0 +1,160 @@
...
+  ActionProtocolClient,
+  ActionProtocolHandler,
+  ActionResponse
+}
+import io.cloudstate.testkit.InterceptService.InterceptorContext
+
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+final class InterceptActionService(context: InterceptorContext) {
```
pvlugter, line 35: Good, was just about to add the intercept for actions.

ralphlaude:
nice work!

protocols/tck/cloudstate/tck/model/eventlogeventing.proto
```diff
@@ -0,0 +1,178 @@
...
+//
+// The `EventLogSubscriberModel` service is an Action that should be implemented in the following ways:
+//
+// - The `ProcessEventOne` method receives an `EventOne` message, and must behave according to the passed in step.
+// - The `ProcessEventTwo` method receives an `EventTwo` message, and must return a stream of responses, one for each
+//   incoming step.
+// - The `Effect` method receives an `EffectRequest` message and must respond with a `Response` that contains the id
+//   from the effect message.
+// - The `ProcessAnyEvent` method receives a `google.protobuf.Any`, which will contain JSON serialized according to
+//   the Cloudstate JSON serialization conventions, with a `type_url` of `json.cloudstate.io/JsonEvent`. The contents
```
marcellanz, line 40: I think the mandated value of `json.cloudstate.io/JsonEvent` here can't be enforced as different languages choose "a path determined using a language-specific mechanism" for the type serialized:
As per https://cloudstate.io/docs/contribute/serialization.html#json-values, 

Although the TCK does not enforce the `type_url`described here:
https://github.com/cloudstateio/cloudstate/commit/b722f866ce962814e7b23efd3c7f69c702148c66#diff-e685c4ba3ccc54f02fab8a427f5f505c63fc8edd56b92dab716e934ccacb2b9aR3295

jroper: The suffix can be anything - but the proxy handles JSON in a particular way, so the TCK does indirectly enforce it.

protocols/tck/cloudstate/tck/model/eventlogeventing.proto
```diff
@@ -0,0 +1,178 @@
...
+  string id = 1 [(.cloudstate.entity_key) = true];
+  string message = 2;
+}
+
+//
+// Each `ProcessStep` is one of:
+//
+// - Reply: reply with the given message in a `Response`.
+// - Forward: forward to another service, in place of replying with a `Response`.
+// - SideEffect: add a side effect to the current reply, forward, or failure.
```
marcellanz, line 142: SideEffect is missing yet in `ProcessStep `.

protocols/tck/cloudstate/tck/model/eventlogeventing.proto
```diff
@@ -0,0 +1,178 @@
...
+}
+
+//
+// The `EventSourcedEntityTwo` service is an event sourced entity that should be implemented in the following ways:
+//
+// - The `EmitJsonMethod` method should emit an event serialised as JSON. This event should:
+//   - Contain a single `message` property with the value of the `message` field in `JsonEvent`.
+//   - Be serialized according to the Cloudstate JSON serialization conventions - that is, with the JSON serialized to
+//     bytes, then placed into a protobuf message with a single bytes field with field number 1.
+//   - Have a type_url of `json.cloudstate.io/JsonEvent`.
```
marcellanz, line 93: The Java Support implementation does return here a `JsonMessage`, instead of a `JsonEvent`as an Event.
Could we write that the type_url has to have a prefix of ``json.cloudstate.io/`?

protocols/tck/cloudstate/tck/model/eventlogeventing.proto
marcellanz: >The suffix can be anything

Then this:
>with a `type_url` of `json.cloudstate.io/JsonEvent`

is inaccurate I think. The suffix is out of reach to be defined by the TCK model implemented. Instead it's what the language support has decided to be for a type of a languages types for the Cloudstate JSON serialisation implementation.

PR #468, opened 2020-10-20
Add TCK model test for Actions

pvlugter:
More on evolving the TCK: #316. Add simple TCK model tests for Actions.

Currently only implemented for java-support, which is set up for TCK model tests. This did uncover a couple of things in the java-support implementation: side effects were not being returned to the proxy, and there was no support for failure replies.

### DIFF
--- a/java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
+++ b/java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2019 Lightbend Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.cloudstate.javasupport.tck.model.action;
+
+import akka.NotUsed;
+import akka.actor.ActorSystem;
+import akka.stream.javadsl.Flow;
+import akka.stream.javadsl.Sink;
+import akka.stream.javadsl.Source;
+import io.cloudstate.javasupport.Context;
+import io.cloudstate.javasupport.ServiceCall;
+import io.cloudstate.javasupport.action.*;
+import io.cloudstate.javasupport.action.Effect;
+import io.cloudstate.tck.model.Action.*;
+import io.cloudstate.tck.model.ActionTwo;
+import java.util.concurrent.CompletionStage;
+
+@Action
+public class ActionTckModelBehavior {
+
+  private final ActorSystem system = ActorSystem.create("ActionTckModel");
+
+  public ActionTckModelBehavior() {}
+
+  @CallHandler
+  public CompletionStage<ActionReply<Response>> processUnary(
+      Request request, ActionContext context) {
+    return Source.single(request).via(responses(context)).runWith(singleResponse(), system);
+  }
+
+  @CallHandler
+  public CompletionStage<ActionReply<Response>> processStreamedIn(
+      Source<Request, NotUsed> requests, ActionContext context) {
+    return requests.via(responses(context)).runWith(singleResponse(), system);
+  }
+
+  @CallHandler
+  public Source<ActionReply<Response>, NotUsed> processStreamedOut(
+      Request request, ActionContext context) {
+    return Source.single(request).via(responses(context));
+  }
+
+  @CallHandler
+  public Source<ActionReply<Response>, NotUsed> processStreamed(
+      Source<Request, NotUsed> requests, ActionContext context) {
+    return requests.via(responses(context));
+  }
+
+  private Flow<Request, ActionReply<Response>, NotUsed> responses(ActionContext context) {
+    return Flow.of(Request.class)
+        .flatMapConcat(request -> Source.from(request.getGroupsList()))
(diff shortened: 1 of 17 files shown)


marcellanz:
>This did uncover a couple of things

nice! Best applied usage of a TCK :)

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
```diff
@@ -0,0 +1,111 @@
...
+import akka.stream.javadsl.Source;
+import io.cloudstate.javasupport.Context;
+import io.cloudstate.javasupport.ServiceCall;
+import io.cloudstate.javasupport.action.*;
+import io.cloudstate.javasupport.action.Effect;
+import io.cloudstate.tck.model.Action.*;
+import java.util.concurrent.CompletionStage;
+
+@Action
+public class ActionTckModelBehavior {
```
marcellanz, line 32: Is this Entity different to the eventsourced TCK entities as it is named `Behaviour`(I think this is an Akka/Actor thing? I don't know Akka) as the eventsourced Entity is not named with this pattern `EventSourcedTckModelEntity`.

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
```diff
@@ -0,0 +1,111 @@
...
+    return Sink.fold(
+        ActionReply.noReply(),
+        (reply, next) ->
+            next.isEmpty() ? reply.withEffects(next.effects()) : next.withEffects(reply.effects()));
+  }
+
+  private ServiceCall serviceTwoRequest(Context context, String id) {
+    return context
+        .serviceCallFactory()
+        .lookup("cloudstate.tck.model.action.ActionTwo", "Call", OtherRequest.class)
```
marcellanz, line 108: No issue here, but I'm surprised to see the Java gRPC service descriptors same as, the ones of the Go prototbuf plugin, do not export typesafe identifiers for services and service names.

marcellanz: … there is work underway to change that for the go gRPC plugin https://github.com/grpc/grpc-go/issues/3762

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
```diff
@@ -0,0 +1,111 @@
...
+import java.util.concurrent.CompletionStage;
+
+@Action
+public class ActionTckModelBehavior {
+
+  private final ActorSystem system = ActorSystem.create("ActionTckModel");
+
+  public ActionTckModelBehavior() {}
+
+  @CallHandler
```
sleipnir, line 38: Just a non-PR comment. Why did we choose CallHandler instead of CommandHandler? I don't understand why we need to change the semantics since the behavior is implicit in the type of Entity used, we shouldn't have to call it CallHandler just because it is a stateless entity. This is definitely the case with "pattern is having no pattern".

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
pvlugter: Actions are not considered to be Entities, in the definitions we've been using. Where an Entity is primarily defined by having a distinct identity, associated with its id or key. I think the usual pattern will be to name actions with `Action`, and I originally had `ActionTckModelAction` and `ActionTwoAction` here, and then tried naming it with `Behaviour` instead to avoid the awkwardness. Happy to find a naming convention and use it here in the TCK too.

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
pvlugter: The Scala service descriptors do have the identifiers. Maybe they've been added for Java as well now. Good that there will be a better API for Go.

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
pvlugter: For Actions or Functions (which are not Entities) I expect `EventHandler` may be the most commonly used in other similar solutions, but I think that was seen as confusing with event-sourced entities having a more specific use of event handling. In some discussions, just using `Handler` for actions was discussed. I'll let @jroper comment on why its `CallHandler` currently. Sticking to `CommandHandler` for both entities and actions seems okay to me too.

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
pvlugter: Akka gRPC for Java does of course generate a bunch of things, including static service names that can be used here. The method descriptors in the client all look to be private, otherwise they could be used for a lookup here. Agree that it would be good to have typesafe APIs for Cloudstate service calls, and we should otherwise look at generating these. The Java support is very reflection and runtime-based currently.

pvlugter: I've updated this to use the generated service name at least, from Akka gRPC (the generation was actually disabled, as only the java protoc generation is needed).

java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/action/ActionTckModelBehavior.java
pvlugter: @sleipnir, do you want to create an issue for the `CallHandler` naming? So we track discussion there and remember to make changes.

pvlugter:
As it's been a couple of weeks, I'll merge this now, and make it simpler for the next TCK extensions. Post-merge reviews still welcome of course.

PR #472, opened 2020-10-29
Retry Spanner schema creation and shutdown on ultimate failure

hseeberger:
See https://github.com/lightbend/cloudstatemachine/issues/2083.

This PR improves on the so far very naive approach to check and create the database schema for Google Cloud Spanner:
- Several admin calls are unified into single a one
- Schema check/creation is retried a configurable number of times after a configurable delay
- After ultimate failure the actor system is shutdown (for k8s to react upon)


pvlugter (approved):
LGTM

PR #479, opened 2020-11-04
Remove state setting from CRDT protocol

pvlugter:
Creating multiple CRDT replicas concurrently would only apply one of the initial states, because of the way the proxy implementation would use the state as the initial value for the underlying CRDT without a modify function for the update. Having the initial state be the empty/zero state plus an initial update will work concurrently. This makes the initial state part of the CRDT protocol redundant, only deltas are needed. The Java support implementation was also buggy — if an empty CRDT was created which then received the initial state from another replica, it would still think it was a new CRDT on updates, returning a create action with another initial state, which would then crash the proxy entity.

So this PR removes the state part of the CRDT protocol, and updates the proxy, java support, and node support implementations to use deltas only. From the proxy, the initial state (whether in init messages or after creation) is always sent as a delta. From the user function, the initial state is generally not sent if the CRDT is still empty/zero. For ORMaps, added values which are empty CRDTs will still send an initial delta of the empty state — this seemed simpler, so that the proxy still receives a typed CRDT that represents the empty initial state.


protocols/protocol/cloudstate/crdt.proto
```diff
@@ -70,25 +69,18 @@ message CrdtStreamIn {
...
-        // does not have a current state, either because the init function didn't send one and the user function hasn't
-        // updated the state itself in response to a command, or because the state was deleted, this must be sent before
-        // any deltas.
-        CrdtState state = 2;
-
-        // A delta to be applied to the current state. May be sent at any time as long as the user function already has
-        // state.
-        CrdtDelta changed = 3;
+        // A delta to be applied to the current state. May be sent at any time.
+        CrdtDelta delta = 2;
```
marcellanz, line 73: nice: changed => delta

jroper (approved):

PR #484, opened 2020-11-07
Allow the proxy to configure passivation for entity type

### LINKED ISSUE #298: Update entity passivation
Akka 2.6 has entity passivation built-in now. See `akka.cluster.sharding.passivate-idle-entity-after` setting and [discussion](https://github.com/cloudstateio/cloudstate/pull/293/files#r419180864).

Or we may want to allow passivation timeouts to be configurable per entity type on the user side?

ralphlaude:
Resolves #298

allow the proxy to configure passivation for entity type.

### DIFF
--- a/proxy/core/src/main/scala/io/cloudstate/proxy/EntityDiscoveryManager.scala
+++ b/proxy/core/src/main/scala/io/cloudstate/proxy/EntityDiscoveryManager.scala
@@ -68,7 +68,8 @@ object EntityDiscoveryManager {
       gracefulTerminationTimeout: Timeout,
       numberOfShards: Int,
       proxyParallelism: Int,
-      journalEnabled: Boolean,
+      eventSourcedSettings: EventSourcedSettings,
+      crdtSettings: CrdtSettings,
       config: Config
   ) {
     validate()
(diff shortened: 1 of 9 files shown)


proxy/core/src/main/scala/io/cloudstate/proxy/EntityDiscoveryManager.scala
```diff
@@ -83,16 +82,21 @@ object EntityDiscoveryManager {
...
            proxyParallelism = config.getInt("proxy-parallelism"),
-           journalEnabled = config.getBoolean("journal-enabled"),
+           journalEnabled = config.getBoolean("eventsourced-entity.journal-enabled"),
            config = config)
     }
 
     final def getConfig(path: String): Config =
       config.getConfig(path)
 
+    final def eventSourcedPassivation(): Timeout =
```
pvlugter, line 94: Rather than a method, I think it would be good to have this as a setting value, like before. We could have an eventsourced settings class, to group these (journal enabled and passivation timeout currently), rather than have them all at the top level.

ralphlaude: Done

pvlugter: Thanks

pvlugter (approved):

Issue #111, opened 2019-09-25
Configurable entity passivation


jroper:
Two things:

* CRDT entities don't currently support passivation, they should.
* The passivation timeout should be configurable. There should be at least two ways to configure it, the first would be in the StatefulService deployment descriptor, the second in the CloudState operator ConfigMap. A third might be to allow it to be configured by the user function itself during entity discovery (this would allow different entity types to have different passivation timeouts). Not sure if that's absolutely necessary or not though.

viktorklang:
@jroper Do you have any idea on how this would be implemented? Is this a configuration job only, or do we need to implement it as well?

jroper:
Basic outline:

* The proxies reference.conf should pick up an entity passivation timeout environment variable, if present (the same way other environment variables are referenced in reference.conf).
* The operator should add this environment variable to the pod. It should look at the stateful service descriptor (which should have the timeout added to it), and if not present there, read it from the configuration that was parsed from the config map. Potentially, it should only set the environment variable if the value differs from the default.

viktorklang:
@jroper I was more thinking about "CRDT entities don't currently support passivation, they should." So is it the case that there is no support for it, or that it is not configurable externally right now?

PR #486, opened 2020-11-13
Passivation configured in language supports via the discovery protocol

### LINKED ISSUE #298: Update entity passivation

ralphlaude:
Resolves #298

Passivation timeouts configured in language supports via the discovery protocol. The first step is the java-support and then node-support.

In the java-support all entity annotations are extended `@XXXEntity (passivationTimeout = 10)`,  the service class `io.cloudstate.javasupport.Service` is also extented with `passivationTimeout` and the grpc entity in the entity.proto.


ralphlaude:
@pvlugter what do you think about the option?

java-support/src/main/java/io/cloudstate/javasupport/crdt/CrdtEntity.java
```diff
@@ -40,4 +40,12 @@
...
 @Retention(RetentionPolicy.RUNTIME)
-public @interface CrdtEntity {}
+public @interface CrdtEntity {
+
+  /**
+   * Specifies the idle time of the entity before it is passivated. Zero and any negative value mean
+   * use default from configuration file. Any positive value means the entity is passivated after
+   * the idle time.
+   */
+  int passivationTimeout() default 0;
```
sleipnir, line 50: Hello Guy I didn't like this option for me it should be done at the Entity registration stage and not at the entity itself.

sleipnir (commented):
Guy thinks that this option should be passed via registryXXXXEntity methods and not in the entity's annotation. I think this pollutes the entity a little.

java-support/src/main/java/io/cloudstate/javasupport/crdt/CrdtEntity.java
ralphlaude: Hi Adriano, it is also fine for me and it is easier. I will do it at the registration stage.

sleipnir: Great! Thanks

ralphlaude: why do you like this option?

marcellanz: Why would 0 not be a valid value? And therefore passivation disabled at all?

ralphlaude: Doing it at the registration stage means breaking the existing registration API and all stay the same with annotations. Then we are forcing user to think about the passivation timeout for the registration. I think the option with annotations is without breaking and easier for the user. @sleipnir I would be glad to hear your idea.

sleipnir: > why do you like this option?

I prefer the registration option because I understand a registration step as a configuration step for the execution service of the entity and not a property of the entity. I also prefer that the entity annotations are as simple as possible, without many options because it seems cleaner. See for example how I make the settings in Kotlin support, the user informs what he wants in the registration step and the entity remains without many changes.

```
fun main() {
    cloudstate {
        eventsourced {
            entityService = ShoppingCartEntity::class
            descriptor = Shoppingcart.getDescriptor().findServiceByName("ShoppingCart")
            additionalDescriptors = mutableListOf(Shoppingcart.getDescriptor(), Domain.getDescriptor() )
           
            // Settings that I want to inform the proxy are here:
            snapshotEvery = 1
            persistenceId = "shopping-cart"
        }

    }.start()
            .toCompletableFuture()
            .get()
}
```

ralphlaude: @marcellanz 0 or negative means use the default config for the passivation.

marcellanz: @ralphlaude yeah saw that documented, but what is the motivation to allow a low time of 1sec, but not 0sec? <0 would be enough to piggyback to take the default config, alas setting to 0 allows the entity to disable or set it to 0.

ralphlaude: > I prefer the registration option because I understand a registration step as a configuration step for the execution service of the entity and not a property of the entity. I also prefer that the entity annotations are as simple as possible, without many options because it seems cleaner. See for example how I make the settings in Kotlin support, the user informs what he wants in the registration step and the entity remains without many changes.

@sleipnir it is just about Java-support here and other languages can use their own ways to do that. I have not idea about Kotlin but I don't see another option in Java. How would you configure the passivation timeout with Kotlin?

sleipnir: > Doing it at the registration stage means breaking the existing registration API and all stay the same with annotations. Then we are forcing user to think about the passivation timeout for the registration. I think the option with annotations is without breaking and easier for the user. @sleipnir I would be glad to hear your idea.

I do not agree. I don't see how adding an option to the registry breaks the current API? Do the values not have defaults? Am I required to set the passivation time? If any of the answers to these questions are true then it would break the registration API, otherwise it would not.

sleipnir: @ralphlaude If the user for any reason wants to use the passivation option then he is already thinking about Passivation, he is not being forced to do anything whether it be done in the annotation or in the registry phase.

marcellanz: I should have written instead of
> setting to 0 allows the entity to disable or set it to 0

"setting to 0 allows the entity passivate immediately."

sleipnir: > > I prefer the registration option because I understand a registration step as a configuration step for the execution service of the entity and not a property of the entity. I also prefer that the entity annotations are as simple as possible, without many options because it seems cleaner. See for example how I make the settings in Kotlin support, the user informs what he wants in the registration step and the entity remains without many changes.
> 
> @sleipnir it is just about Java-support here and other languages can use their own ways to do that. I have not idea about Kotlin but I don't see another option in Java. How would you configure the passivation timeout with Kotlin?

I know that we are talking about java support, I just gave you an example to explain the concept of what I was telling you. If you don't think it is necessary to change then ok, I just presented my arguments.

ralphlaude: @marcellanz the entity should not passivate immediately and it should be allowed. Setting to 0 allows the entity to passivate after 30 seconds. The`passivateTimeout` in the annotation should define the minimal value to accept.

sleipnir: > @marcellanz the entity should not passivate immediately and it should be allowed. Setting to 0 allows the entity to passivate after 30 seconds. The`passivateTimeout` in the annotation should define the minimal value to accept.

This seems counter intuitive to me and should be well documented if it is assumed. Setting a property to 0 and assuming a value of 30 is somewhat unusual. If this is really the case, we cannot forget to be very clear about this in the documentation.

ralphlaude: @sleipnir passivation is always enabled and is implicit to the user. The `registerXXXEntity` does not  force the user to think about passivation. The new `registerXXXEntity(..., passivationTimeout)` forces the user to provide passivation timeout or at least to think about it.
For now i see two options here:
1- Annotations
2- Provide overloaded registration methods with passivationTimeout

sleipnir: > @sleipnir passivation is always enabled and is implicit to the user. The `registerXXXEntity` does not force the user to think about passivation. The new `registerXXXEntity(..., passivationTimeout)` forces the user to provide passivation timeout or at least to think about it.
> For now i see two options here:
> 1- Annotations
> 2- Provide overloaded registration methods with passivationTimeout

Guy Registry phase doesn't force anyone, I already told you, sorry your argument doesn't convince me, I'm sorry. As I told you before, if you don't feel comfortable don't change, I made my contribution to the topic and I will not argue anymore.

ralphlaude: > I know that we are talking about java support, I just gave you an example to explain the concept of what I was telling you. If you don't think it is necessary to change then ok, I just presented my arguments.

understood and i saw how it works in Kotlin

ralphlaude: > This seems counter intuitive to me and should be well documented if it is assumed. Setting a property to 0 and assuming a value of 30 is somewhat unusual. If this is really the case, we cannot forget to be very clear about this in the documentation.

@sleipnir I agree with default values in annotations and it already works that way with attributes (persistentId and snapshotEvery) in `@XXXEntity`. It is also well documented. May be we should change that and explicitly work with default values in annotations.

pvlugter: I agree generally about keeping the entity implementations simpler by not having too many options on the annotations. Also understand that this is the current approach in Java support, and annotation settings allow for defaults. So I'd be inclined to have both annotation and registration options right now. The registration method for annotated entity classes doesn't currently have extra options, but it makes sense to add them now. Not sure how many more options there will be in the future, but I'd consider having a settings object (so something like `EventSourcedOptions` or similar), instead of supporting multiple permutations of parameters — so the method can just be overloaded with and without the options, and so that the options object can have default or unset options.

Regarding default of 0, I'd probably interpret a timeout of 0 as disabling passivation altogether. Not sure if we would want to support disabling passivation, where the user function would control lifecycle itself or something, but I think I'd avoid 0 as being used for an unset option for selecting the proxy default.

I think it makes sense to improve the APIs and protocol approaches whenever possible. For a timeout setting, it would be good to support any kind of duration (not just int of seconds). So I would consider accepting `java.time.Duration` in the java support APIs, or `TimeUnit` variations. Annotation values are limited, can't use arbitrary classes, so one option instead of an int is to use a String and use the typesafe Config duration parsing or `Duration.parse` (which is a bit different). And can then use empty String as having the timeout unset.

For the settings that we want to have fall back to proxy-side defaults, I would support this directly in the settings — so the entity options object knows when a timeout has been set or not, and same in the protocol — and it only sends a timeout if it's been configured in the user function, rather than relying on negative numbers. And the protocol could have a duration message, with time units, to make things complete.

So I would aim for something along the lines of:

```java
// no passivation timeout set, nothing sent with entity spec to proxy
@EventSourcedEntity(persistenceId = "something")
```

```java
// passivation timeout parsed from string, sent as duration message in protocol
@EventSourcedEntity(persistenceId = "something", passivationTimeout = "10s")
```

```java
// no extra options specified, use annotation options or defaults
.registerEventSourcedEntity(
  SomeEntity.class,
  SomeProtocol.getDescriptor().findServiceByName("SomeService"),
  SomeProtocol.getDescriptor())
```

```java
// options specified, which will override annotation options or defaults
.registerEventSourcedEntity(
  SomeEntity.class,
  SomeProtocol.getDescriptor().findServiceByName("SomeService"),
  SomeProtocol.getDescriptor(),
  EventSourcedOptions.defaults().setPassivationTimeout(Duration.ofSeconds(10))
```

sleipnir: I agree with the whole proposal here

ralphlaude: After the discussion last week, i agree fully with the proposal.

marcellanz: >and it only sends a timeout if it's been configured in the user function, rather than relying on negative numbers. And the protocol could have a duration message, with time units, to make things complete.

I agree to change the protocol now and a soon as possible for such changes. I also agree on not depending on magic values.

jroper: I think I'd like to have options for expression the following things - perhaps not immediately, but we should choose a mechanism and protocol that allows us to add support for them in future if needed:

* A specific passivation timeout, with at least millisecond resolution.
* Default - note that I think it's fine for the default to come from the support library, I don't see any reason why the proxy needs to be in control of the default.
* No passivation timeout - entities should never passivate unless rebalanced.
* Immediate passivation, this effectively means the state is reloaded for each command.
* Some way to indicate the proxy should apply smarter, more dynamic passivation strategies, for example strategies that respond to resource constraints, or perhaps auto-tune themselves based on cardinality and/or hot spot metrics.

So, basically, I think it's worth now choosing a strategy in the APIs and protocol that will allow us to add some of the features above without breaking the protocol or APIs.

pvlugter: Good idea to future-proof the protocol and APIs now for other passivation options.

ralphlaude: Really good idea for passivation options.
More generally we should have a strategy for adding other future options (not only passivation) without breaking the protocol or the APIs.

pvlugter: After some more discussion:

- passivation timeout in protocol can just be milliseconds, we don't expect more fine-grained than that
- language supports can use duration APIs or similar as appropriate, convert to milliseconds
- protocol should support expanding to other passivation strategies, where initially it could just be `timeout`, ideally we can add `never` or `immediate` or other strategies in a protocol-compatible way
- protocol should support strategies explicitly, such as `timeout`, `never` or `immediate`, and not use magic values

ralphlaude:
The discovery process, the proxy and the API have been extended to support passivation strategy and also that new passivation strategy (`never`, `immedaite` and perhaps resource specifics strategies) can be added without breaking the API. The only supported passivation strategy right now is `timeout`. Passivation strategy could be as complex as possible to configure and this does not fit within annotations. So the approach to pass configuration parameter at registration time is a better fit. There is `EntityOptions` for configuring passivation strategy and other aspects if needed.
The language support should deal with default passivation timeout and more generally fallback.
All entities must support a passavation strategy and right now it is `timeout`. The entity.proto file has been modified like this:
```
message Entity {
    ...

    // The passivation strategy for the entity.
    EntityPassivationStrategy passivation_strategy = 4;
}

// A passivation strategy for the entity which is sent to the proxy.
message EntityPassivationStrategy {
    oneof strategy {
	// the timeout passivation strategy.
        TimeoutPassivationStrategy timeout = 1;
        ImmediatePassivationStrategy immediate = 2; // this strategy can be added later on
        NeverPassivationStrategy never = 3; // this strategy can also be added later on
        // more strategies can be added here
    }
}

// A passivation strategy based on a timeout. The idle timeout after which an entity is passivated.
message TimeoutPassivationStrategy {
    // The timeout in millis
    int64 timeout = 1;
}

// A passivation strategy which passivates the entity after a command is handled.
message ImmediatePassivationStrategy {}

// A passivation strategy which never passivates the entity.
message NeverPassivationStrategy {}
```

For the java support it will be like at the end:
```java

// passivation stratgey timeout with default passivation timeout set which is sent with entity spec to proxy
@EventSourcedEntity(persistenceId = "something")

// no extra options specified, passivation stratgey timeout with default timeout sent with entity spec to proxy
.registerEventSourcedEntity(
  SomeEntity.class,
  SomeProtocol.getDescriptor().findServiceByName("SomeService"),
  SomeProtocol.getDescriptor())

// options specified with passivation stratgey timeout, which will be sent with entity spec to proxy
.registerEventSourcedEntity(
  SomeEntity.class,
  SomeProtocol.getDescriptor().findServiceByName("SomeService"),
  EventSourcedEntityOptions.defaults().withPassivationStrategy(PassivationStrategy.timeout(Duration.ofSeconds(10)))
  SomeProtocol.getDescriptor())
```

Only the java support is implemented and all others language supports (Nodejs, GO, Kotlin, Python, Spring and so on) should be extended before this can be merged.
Wo can help with other language supports (Nodejs, GO, Kotlin, Python, Spring and so on)?

marcellanz:
Thanks @ralphlaude :) For Go for sure I can help.
If needed we could work based on a dedicated branch so that maintained language supports can be aligned. As the new field `EntityPassivationStrategy` protocol-wise is an addition, entity discovery could probably even work without the support on user languages?

ralphlaude:
@marcellanz thanks for Go.
Yes we can think how to work on a dedicated branch to be aligned.
With the addition of the new field `EntityPassivationStrategy` in the protocol the discovery is broken without the support on user language. It can be changed if we want it.

sleipnir:
Hi @ralphlaude , thanks for another one!
I can help with the other languages mentioned but generally as Kotlin and Springboot depend on the java-support jar I choose to wait for the release of an official release with the functionality, I do this to not have to build the jars locally. I sincerely wish there were more releases to help me with this, but I may have to follow the branches of the main repository with equivalent branches in the languages, I don't know, I find this approach very laborious, but this should be discussed in another forum.

marcellanz:
@sleipnir I think, it should be possible or better said desirable for language supports to evolve before the core project releases. For non-JVM supports it is already manageable with tck- and proxy images available from master. For JVM support perhaps snapshot releases help? But as you said, could be discussed elsewhere.

sleipnir:
> For JVM support perhaps snapshot releases help? ...

Yes!

pvlugter:
> > For JVM support perhaps snapshot releases help? ...
> 
> Yes! But they are also often not available.

It's easy to publish a versioned snapshot for java support. Just ask if you need one :)

pvlugter:
> With the addition of the new field `EntityPassivationStrategy` in the protocol the discovery is broken without the support on user language. It can be changed if we want it.

Since it's an additional field, the protocol can be compatible with older versions. In the proxy, instead of throwing exceptions if the language support doesn't define a passivation strategy, we can use the proxy-side default.

ralphlaude:
> Since it's an additional field, the protocol can be compatible with older versions. In the proxy, instead of throwing exceptions if the language support doesn't define a passivation strategy, we can use the proxy-side default.
> 
> I think it will be better to have this compatible and use the proxy default, rather than require all language supports to be updated straight away. There's also an older issue on configurable entity passivation (#111) which describes having the passivation strategy defined in the Kubernetes CRD or ConfigMap and injected by the operator — which would allow the proxy-side default to be configured, if the language support doesn't define it.

The proxy-default for passivation strategy will be introduced again to allow the compatibility so we can move forward. I think it would better in the middle term to have a central place where to configure things like passivation strategy. Now we have two places where passivation strategy is configured. I think the better place for that is the language support.

ralphlaude:
Everything is done here. I am not sure if we want to extends the nodejs support here or do it in another issue. If yes, I will create issues for nodejs support and another language support (GO, Spring, Kotlin, Python and others).
@sleipnir, @marcellanz what do you think?

marcellanz:
@ralphlaude I think for changes on the protocol one implementation for a "main" user support language should be enough. With "main" this might be JS or Java I think.

As the protocol changed more often lately, it reveals how we might work with such changes in general. With changes worked on concurrently, having the proxy and TCK in the right state for development leads to a bit of coordination. 
I think we don't need overly complex rules for that like, as an example, opening issues for all support language. Perhaps the TCK can help to guide where changes are pending for other support languages. This spans a matrix of combinations for proxy, TCK and support language implementations. The current situation already shows this, with currently a few languages aligned with master, and others partly up to date.

This said, I think support languages should aim to get protocol updates by themselves. If the protocol changes by a PR like this one, the changing PR a) should provide a TCK test or if not applicable, b) a comment in the spec about semantics changed in the protocol.

The TCK Verification here https://observablehq.com/@marcellanz/cloudstate-tck-verification displays the current situation also in an automatic way, with pending TCK tests visible.

WDYT?

sleipnir:
@ralphlaude  I agree with @marcellanz  but I add that the comment on the semantics desired by the protocol must be mandatory in PR. This avoids that developers have to have a long journey of understanding evaluating codes and contracts where a comment would be enough to make the intention clear.

ralphlaude:
@marcellanz the proposal is fine for me. It would be nice to have support for nodejs. I will comment the spec about the new semantic.

ralphlaude:
@pvlugter, @jroper I need support here for nodejs. Please take a look.

java-support/src/main/scala/io/cloudstate/javasupport/impl/PassivationStrategies.scala
```diff
@@ -0,0 +1,34 @@
...
+
+import com.typesafe.config.ConfigFactory
+import io.cloudstate.javasupport.PassivationStrategy
+
+private[impl] case class Timeout(duration: Duration) extends PassivationStrategy
+
+private[impl] object Timeout {
+
+  private val defaultPassivationTimeout: Duration = {
+    val config = ConfigFactory.load()
```
pvlugter, line 29: Ideally the config is not loaded in multiple places (even though the config loading is cached), so that it's only defined once, especially if custom config can be passed into the top-level classes. Looks like `CloudStateRunner` has a constructor that accepts a config. Could we get move the defaults to one place, so they can all use the same config?

protocols/protocol/cloudstate/entity.proto
```diff
@@ -262,6 +262,29 @@ message Entity {
...
+
+    // The passivation strategy for the entity.
+    EntityPassivationStrategy passivation_strategy = 4;
+}
+
+//
+// The semantics is to provide a flexible way for entity user functions to configure the passivation strategy.
+// This strategy is sent to the proxy at discovery time allowing the proxy to configure the corresponding entities.
+// The only passivation strategy supported is the timeout strategy and configuring this is optional for the entity.
+// If an entity user function does not configure the passivation strategy the proxy used its fallback default value.
```
pvlugter, line 274:
```suggestion
// If an entity user function does not configure the passivation strategy the proxy uses its fallback default value.
```

protocols/protocol/cloudstate/entity.proto
```diff
@@ -262,6 +262,29 @@ message Entity {
...
+//
+// The passivation strategy for the entity user function.
+message EntityPassivationStrategy {
+    oneof strategy {
+        // the timeout passivation strategy.
+        TimeoutPassivationStrategy timeout = 1;
+    }
+}
+
+// A passivation strategy based on a timeout. The idle timeout after which an user function's entity is passivated.
```
pvlugter, line 284:
```suggestion
// A passivation strategy based on a timeout. The idle timeout after which a user function's entity is passivated.
```

proxy/core/src/main/scala/io/cloudstate/proxy/crdt/CrdtSupportFactory.scala
```diff
@@ -93,9 +94,10 @@ class CrdtSupportFactory(system: ActorSystem,
     val methodsWithoutKeys = methodDescriptors.values.filter(_.keyFieldsCount < 1)
     if (methodsWithoutKeys.nonEmpty) {
       val offendingMethods = methodsWithoutKeys.map(_.method.getName).mkString(",")
-      throw new EntityDiscoveryException(
-        s"CRDT entities do not support methods whose parameters do not have at least one field marked as entity_key, " +
-        "but ${serviceDescriptor.getFullName} has the following methods without keys: ${offendingMethods}"
+      throw EntityDiscoveryException(
+        s"""CRDT entities do not support methods whose parameters do not have at least one field marked as entity_key,
+            |but ${serviceDescriptor.getFullName} has the following methods without keys: ${offendingMethods}""".stripMargin
+          .replaceAll("\n", " ")
```
pvlugter, line 100: Is this to fix the string interpolation missing from the second string? Rather than multiline string, with margin stripped, and then newlines replaced, could just add string interpolation to the second string (`s"`) and not the first one.

proxy/core/src/main/scala/io/cloudstate/proxy/eventsourced/EventSourcedSupportFactory.scala
```diff
@@ -83,9 +86,10 @@ class EventSourcedSupportFactory(
     val methodsWithoutKeys = methodDescriptors.values.filter(_.keyFieldsCount < 1)
     if (methodsWithoutKeys.nonEmpty) {
       val offendingMethods = methodsWithoutKeys.map(_.method.getName).mkString(",")
-      throw new EntityDiscoveryException(
-        s"Event sourced entities do not support methods whose parameters do not have at least one field marked as entity_key, " +
-        "but ${serviceDescriptor.getFullName} has the following methods without keys: ${offendingMethods}"
+      throw EntityDiscoveryException(
+        s"""Event sourced entities do not support methods whose parameters do not have at least one field marked as entity_key,
+            |but ${serviceDescriptor.getFullName} has the following methods without keys: ${offendingMethods}""".stripMargin
+          .replaceAll("\n", " ")
```
pvlugter, line 92: Same here, if it's to fix the missing string interpolation, just move the `s"` to the second string.

pvlugter (commented):
Looking good. Added a few review comments — would be good to not load the config in the Timeout static object in particular, but have the default come from the top-level config.

pvlugter:
Yes, would be good to add passivation timeouts for node support. Node support is behind the Java support in a few areas now: applying events immediately (and any failure handling for this) #375, support for value entities, new TCK implementation, and this. Doesn't need to be added to this PR, let's create an issue to track.

pvlugter:
Could be useful to have passivation timeouts as part of the TCK for testing.

java-support/src/main/scala/io/cloudstate/javasupport/impl/PassivationStrategies.scala
ralphlaude: I will take a look and see how the issue can be solved.

ralphlaude: > Could we get move the defaults to one place, so they can all use the same config?

@pvlugter I am not sure I understand. Could you please be more specific?

ralphlaude:
> Could be useful to have passivation timeouts as part of the TCK for testing.

@pvlugter Could you please be more specific?

java-support/src/main/scala/io/cloudstate/javasupport/CloudStateRunner.scala
```diff
@@ -79,15 +79,16 @@ final class CloudStateRunner private[this] (
...
-  // TODO JavaDoc
+  /**
+   * Creates a CloudStateRunner from the given services. Use the default config to create the internal ActorSystem.
+   */
   def this(services: java.util.Map[String, java.util.function.Function[ActorSystem, Service]]) {
-    this(ActorSystem("StatefulService", {
-      val conf = ConfigFactory.load()
-      conf.getConfig("cloudstate.system").withFallback(conf)
-    }), services.asScala.toMap)
+    this(ActorSystem("StatefulService", CloudStateConfigHolder.defaultConfiguration()), services.asScala.toMap)
```
pvlugter, line 86: It's the other constructor, that takes a config explicitly, that should be supported. The config gets passed to the actor system. It should also be the config that gets passed through to other places (could be retrieved from the actor system). Might be more obvious if there's a test that sets the passivation timeout in config?

pvlugter:
> > Could be useful to have passivation timeouts as part of the TCK for testing.
> 
> @pvlugter Could you please be more specific?

Since passivation strategy is part of the protocol, it would be useful to test it in the TCK. So testing that language supports can configure this, and ideally also testing that the proxy applies the timeout and passivates the entity. For the TCK, it would be fine to have an additional entity defined which is just for testing this, since the passivation timeout is on entity discovery.

java-support/src/main/scala/io/cloudstate/javasupport/CloudStateRunner.scala
ralphlaude: @pvlugter I hope my solution does the job and I did not find a better way.

ralphlaude:
TCK tests for passivation has been added which uses value-based entities.

tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala
```diff
@@ -172,38 +180,85 @@ class CloudStateTCK(description: String, settings: CloudStateTCK.Settings)
           serviceNames must contain("EventSourcedTckModel")
           entity.entityType mustBe EventSourced.name
           entity.persistenceId mustBe "event-sourced-tck-model"
+          entity.passivationStrategy must (not be None or be(None))
```
pvlugter, line 183: Don't think it makes sense to include a check for none or not none :)

tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala
```diff
@@ -172,38 +180,85 @@ class CloudStateTCK(description: String, settings: CloudStateTCK.Settings)
...
+      import io.cloudstate.tck.model.entitypassivation.entitypassivation._
+
+      var entityId: Int = 0
+      def nextEntityId(): String = { entityId += 1; s"entity:$entityId" }
+
+      def passivationEntityTest(test: String => Any): Unit =
+        testFor(PassivationTckModel)(test(nextEntityId()))
+
+      def passivationTimeout(entity: Entity): FiniteDuration = {
+        val additionalTimeout = 100 // additional time for allowing passivation
```
pvlugter, line 236: May need to be longer for CI tests

pvlugter (commented):
Okay, I can always follow up with something that doesn't use a global static config

tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala
```diff
@@ -169,52 +169,54 @@ class CloudStateTCK(description: String, settings: CloudStateTCK.Settings)
...
-          entity.passivationStrategy must (not be None or be(None))
+          entity.passivationStrategy must not be None
         }
 
         spec.entities.find(_.serviceName == EventSourcedShoppingCart.name).foreach { entity =>
           serviceNames must contain("ShoppingCart")
           entity.entityType mustBe EventSourced.name
           entity.persistenceId must not be empty
-          entity.passivationStrategy must (not be None or be(None))
+        //entity.passivationStrategy must not be None // some language support does not support passivation yet!!
```
marcellanz, line 198: This would then show this test "green" for language supports not implementing passivation Strategy? 

If yes, perhaps this should show a failed test. The TCK should show which language support libraries are up-to-date. The TCK will probably show for some time a lot of "red and orange" for other implementations
https://observablehq.com/@marcellanz/cloudstate-tck-verification as the protocol progresses.

proxy/core/src/main/resources/reference.conf
```diff
@@ -147,7 +147,7 @@ cloudstate.proxy {
         enabled = false
 
         # Passivation timeout for Value Entities
-        passivation-timeout = 30s
+        passivation-timeout = 30s // Keep short for testing purposes
```
marcellanz, line 150: Are 30s then also a good option for production?
Perhaps we might document that, that the default is not the value to be chosen for production environments.

ralphlaude: It is true and it will be documented. I am not sure if 30s is a good option for production. Perhaps @pvlugter can give more insight here.

tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala
ralphlaude: Good point. I will rather try to put the test on pending for language support not implementing the passivation strategy

ralphlaude:
@pvlugter ping! you can follow up.

pvlugter:
I'll just close and reopen this PR to see if we have Travis CI working again.

pvlugter (approved):
👍🏼 Good to see the global config part updated.

I'll merge this now, and include in the TCK reorganisation.

PR #491, opened 2020-11-19
Add TCK model test for CRDT entities

pvlugter:
Add tests for CRDT entities to the TCK. Currently only implemented for Java support, with some small changes: tweaks to LWWRegister, and support for configuring write consistencies.

The TCK test class is getting quite long now. Will follow-up with some reorganisation to break this into multiple files.

And will follow-up with the rename of CRDT Entity to Replicated Entity once this is merged.


pvlugter:
I'll merge this now, so I can continue building on it for next steps. @marcellanz, just let me know if there are any issues from implementing this, and removing the state setting part of the CRDT protocol.

marcellanz:
@pvlugter thanks for the excellent work on this 👍

I'm down to a few failing test cases and found some bugs and implementation-specific differences too. Great to see the CRDT protocol verified by the TCK to this extent.

PR #492, opened 2020-11-20
Update value entity TCK to require sorted sequences

### LINKED ISSUE #490: TCK: order of state action updates for addItem by the value entity ShoppingCart
The `com.example.valueentity.shoppingcart.ShoppingCart` implements `addItem` in a way that returns state action updates in the order so that the updated line-item, if it has existed, is positioned at the end of the state action updates list:

This is originated because a shopping cart line-item that already exists is removed and then re-added with the newly updated quantity:

https://github.com/cloudstateio/cloudstate/blob/master/samples/java-shopping-cart/src/main/java/io/cloudstate/samples/shoppingcart/ShoppingCartEntity.java#L58-L59
```
    Domain.Cart cart = ctx.getState().orElse(Domain.Cart.newBuilder().build());
    Domain.LineItem lineItem = updateItem(item, cart);
    List<Domain.LineItem> lineItems = removeItemByProductId(cart, item.getProductId());
    ctx.updateState(Domain.Cart.newBuilder().addAllItems(lineItems).addItems(lineItem).build());
```

So adding 2 times `product:2` then 11 of `product:1` and then 31 of `product:2` emits the following state as `ValueEntityUpdate`:

```
command: entity_id:"cart:1" id:3 name:"AddItem" payload:{[type.googleapis.com/com.example.valueentity.shoppingcart.AddLineItem]:{user_id:"cart:1" product_id:"product:2" name:"Product2" quantity:2}}, state: [type.googleapis.com/com.example.valueentity.shoppingcart.persistence.Cart]:{items:{productId:"product:1" name:"Product1" quantity:1} items:{productId:"product:2" name:"Product2" quantity:2}}
command: entity_id:"cart:1" id:4 name:"AddItem" payload:{[type.googleapis.com/com.example.valueentity.shoppingcart.AddLineItem]:{user_id:"cart:1" product_id:"product:1" name:"Product1" quantity:11}}, state: [type.googleapis.com/com.example.valueentity.shoppingcart.persistence.Cart]:{items:{productId:"product:2" name:"Product2" quantity:2} items:{productId:"product:1" name:"Product1" quantity:12}}
command: entity_id:"cart:1" id:5 name:"AddItem" payload:{[type.googleapis.com/com.example.valueentity.shoppingcart.AddLineItem]:{user_id:"cart:1" product_id:"product:2" name:"Product2" quantity:31}}, state: [type.googleapis.com/com.example.valueentity.shoppingcart.persistence.Cart]:{items:{productId:"product:1" name:"Product1" quantity:12} items:{productId:"product:2" name:"Product2" quantity:33}}
```

with the `items:` ordered as shown. This results in the TCK testing this order by the specifics of this implementations detail:
https://github.com/cloudstateio/cloudstate/blob/master/tck/src/main/scala/io/cloudstate/tck/CloudStateTCK.scala#L1390-L1412

We could 
- mandate the TCK `addItem` test case for the shopping cart to be mandated like this
- allow the TCK to match list items independently of their order
- mandate that the order of the carts state for `addItem`, and any other, is the order in which the items where added and or removed.

I think I would opt-in for the last option, whereas the second option also could be helpful for certain instances, where the order is not important at all and would not fit a languages idiomatic implementation.

pvlugter:
Resolves #490. Update the value entity shopping cart to sort items by product id in responses. Allows the TCK tests to have an expected order. We can also update the TCK to test these with any order, which would mean parsing the json strings too. But we're planning to remove shopping cart from the TCK and just use model-based tests anyway.

### DIFF
--- a/samples/java-shopping-cart/src/main/java/io/cloudstate/samples/shoppingcart/ShoppingCartEntity.java
+++ b/samples/java-shopping-cart/src/main/java/io/cloudstate/samples/shoppingcart/ShoppingCartEntity.java
@@ -24,6 +24,7 @@ import io.cloudstate.javasupport.entity.CommandContext;
 import io.cloudstate.javasupport.entity.CommandHandler;
 import io.cloudstate.javasupport.entity.Entity;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
@@ -43,7 +44,10 @@ public class ShoppingCartEntity {
   public Shoppingcart.Cart getCart(CommandContext<Domain.Cart> ctx) {
     Domain.Cart cart = ctx.getState().orElse(Domain.Cart.newBuilder().build());
     List<Shoppingcart.LineItem> allItems =
-        cart.getItemsList().stream().map(this::convert).collect(Collectors.toList());
+        cart.getItemsList().stream()
+            .map(this::convert)
+            .sorted(Comparator.comparing(Shoppingcart.LineItem::getProductId))
+            .collect(Collectors.toList());
     return Shoppingcart.Cart.newBuilder().addAllItems(allItems).build();
   }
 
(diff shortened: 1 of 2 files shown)


marcellanz:
Cool, thanks.
Good to see the shopping cart removed from the TCK.

ralphlaude:
Thanks for fixing the order in the value entity shopping cart.

PR #494, opened 2020-11-25
Fix empty streamed responses (for CRDT entities)

### LINKED ISSUE #493: Empty streamed responses are not actually connected
The presence part of the chat sample uses stream connection and cancellation and a Vote CRDT to track online presence. This is currently not working: cancelling a call to the `connect` method doesn't cause a stream cancel event. This is since support for metadata was added (#366).

The metadata support for streamed methods uses the first streamed response message to extract the metadata to put into response headers, and then connects the response stream. The `connect` method in the presence sample doesn't send any messages, so it doesn't connect the stream, so it can't be cancelled.

A workaround to get the chat example working is to add a `return {};` to the end of the `connect` method, so that it sends an initial empty response and is connected.

Given that we're putting headers in the response and these are extracted from the first message, at the moment we can't also connect a possibly empty stream body to the response, as we can't return a response until we have the metadata for headers. So we now have a new requirement that streamed responses at least send an initial message to be connected.

We should either update the samples and document this (for the next released version of Cloudstate) or change the approach to support empty streams again.

pvlugter:
Resolves #493. Fix for this was simple: don't filter out empty client actions in `UserFunctionRouter`. We need a first empty response to always be sent downstream, otherwise an HttpResponse is never created due to waiting for the first response to extract metadata. Empty client actions will be filtered out in `CommandHandler.processReplies` instead.

Added tests to the TCK for this, including a proxy intercept test that has same behaviour as the presence example but using the CRDT TCK model entity.

@marcellanz, some updates to the CRDT TCK to include for Go support.

### DIFF
--- a/java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/crdt/CrdtTckModelEntity.java
+++ b/java-support/tck/src/main/java/io/cloudstate/javasupport/tck/model/crdt/CrdtTckModelEntity.java
@@ -131,7 +131,7 @@ public class CrdtTckModelEntity {
   }
 
   @CommandHandler
-  public Response processStreamed(
+  public Optional<Response> processStreamed(
       StreamedRequest request, StreamedCommandContext<Response> context) {
     if (context.isStreamed()) {
       context.onChange(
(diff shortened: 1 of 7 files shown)


marcellanz:
>@marcellanz, some updates to the CRDT TCK to include for Go support.

Thanks @pvlugter. I'll add them with https://github.com/cloudstateio/go-support/pull/56 when this PR is merged I think.

jroper (approved):

PR #497, opened 2020-12-03
Fix proto descriptor extract service

### LINKED ISSUE #496: Failure: multiple grpc services in the same package
Reported bug: registering multiple entities where the grpc services are in separate proto files but in the same package will currently fail the proxy with: `IllegalArgumentException: requirement failed: ServiceDescriptor not found, please verify the spelling and package name provided when looking it up`.

pvlugter:
Resolves #496.

If there are multiple entities registered, which have separate file descriptors but share the same package name, then subsequent services can't be found. The `extractService` method is used to collect the first matching service. Previously it would filter on the package, and then do a `findServiceByName`, but this will return a `Some(null)` on an earlier file descriptor if the service is actually in a later file descriptor. Update to only return `Some(serviceDescriptor)` if found, otherwise `None` to continue the search. I'm assuming this was the original intention.

Can follow up with a test...


jroper:
I think I fixed this in #465.

pvlugter:
> I think I fixed this in #465.

:+1: Ah, didn't think of checking that PR. Yes, fixed in the same way with a `flatMap` (just unnecessary option allocations).

pvlugter:
I'm assuming #465 ends up testing this in some way then, if it was uncovered then.

PR #500, opened 2020-12-17
Reorganise and extend TCK

### LINKED ISSUE #316: Evolve the TCK
After a short Gitter discussion with @pvlugter and @sleipnir, I open this issue for discussion (attached is the gitter exchanges for context).

In short, the issue is to evolve the TCK so that a language support library implementation can be verified.

I think the topic is too broad to capture it in one issue and I propose to discuss it here and then create actionable issues out of it. I list here suggested topics and existing related issues.

**Suggested Objectives**
1. Evolve the TCK for all state models defined by the Cloudstate Protocol.
1. Enhance or adapt the TCK Shopping Cart Example to include other state models as explanatory as possible. This might not be exhaustive in terms of TCK completeness, but still have a use case that is easy to understand, and also can be an example how to use the state models. I'm not sure if this is that important for the TCK.
1. Define synthetic entities and interactions so that the Cloudstate Protocol can be verified by the TCK. These entities have to be provided then by each support library so that the TCK can run against them.
1. Let support libraries execute the TCK within their repositories, preferable without the need of a JVM to be installed or a SBT dependency => a Docker based approach might be feasible.
1. Maybe, existing [tests of the proxy](https://github.com/cloudstateio/cloudstate/tree/v0.5.0/proxy/core/src/test/scala/io/cloudstate/proxy/crdt) can be used to guide the synthetic tests defined.
1. Perhaps for the future and a low priority for now: the TCK currently runs as "Akka + <name_of_language>'"; how would the TCK test a reference implementation of the proxy.

**Related Issues**
- #216 – Make the TCK easily executed by different language supports
- #289 – Language support module release cycles

---
**Related discussion** (Gitter, May 11th)
> @marcellanz : I think it may be time to extend the TCK for other state models like CRDTs, Stateless, KV, Eventing, <name_any> for Cloudstate. I come to a point where I'd like to verify an implementation and can't. The TCK so far was a great enabler getting started with EventSourcing.
> 
> There is tck/src/ with the Shopping Cart example. How should we model the TCK around other state models? While the proxy itself has many synthetic test cases (AFAIKS) the TCK shopping cart is a nice and illustrative example, should we find such an example for other types too, or perhaps expand on the shopping cart?
> With a few language support libraries in different states existing, and new coming, we also might better have the TCK before they are complete.

>@sleipnir: I think we should do a lot of things in several steps (parallel or not) at TCK. I have illustrated this need several times and I agree with you. So my thinking is divided like this?
> 1. The TCK must be less than a selected docker image and accept the necessary parameters to run the tests easily, without depending on the entire Cloudstate project.
> 1. We must expand the TCK and the shopping cart example, there is a reference somewhere in the list that can be used (HotItems) for example to test "forwards and effects" that are not currently supported in the TCK.
> 1. We must include (with shoppingcart or another) a test for statelessness (a simple request/response would be enough in this case).
> 1. A basic example for CRDT should be included at first and then expanded to more complex examples. I also think it would be a good way to test the various types of CRDT independently, that way we could have a compatibility matrix supporting in supported idioms more granular.

>@pvlugter: Yes, please open an issue, if it doesn't exist already, with everything we need for the TCK. And we have cloudstateio/cloudstate#216 for supporting the TCK in the language supports.
> I'm personally not sure whether having the TCK based on examples, like shopping cart, or just artificial tests is better. Possibly a combination in the end? Simple synthetic test cases to get full coverage, plus examples that demonstrate the main use cases.

pvlugter:
The last part to resolve #316. Reorganise the TCK into separate files, rather than one very long file. Add proxy tests with interceptor based on the TCK model implementations rather than shopping cart, testing gRPC and HTTP APIs. Remove the shopping cart, and disable the unimplemented TCKs, just java-support for now.

There are some things I'll follow up with after this too:
- [x] Now that shopping cart is not used in the TCK, we should switch to requiring protocol version to be reported by the language support and checked for compatibility
- [x] The passivation test for CRDT entities is marked `pendingUntilFixed` — the stream is not completed gracefully but fails with a peer closed exception
- [x] Try with the latest Go support TCK implementation


pvlugter:
Updated the Go support and TCK implementation for most of the new tests in this PR and using a new `cloudstate-go-tck:latest` image in the integration tests now. https://github.com/cloudstateio/go-support/pull/67

pvlugter:
Added the follow-ups directly to this PR:

Fixed the stream completion on passivation for CRDT entities — this was similar to a previous issue for event sourced entities, where the actor was being stopped without first completing the stream, and the stream would then fail and close ungracefully when the stream sees the actor terminate, and would also log a deadletter for the failure since the actor is already stopped; fixed by closing the stream and waiting for it to be closed before stopping the actor. 

Updated the testkit interceptor expect methods from `expectClient`/`expectService` to `expectIncoming`/`expectOutgoing` as I think just using incoming/outgoing messages could be clearer than client/service, with all the other uses of client and service.

Compatible protocol version reported by the language support is now required. The Java and Go supports already do this, and are currently the only language supports tested with the TCK integration tests. There's a setting to disable the check in the proxy if needed, since the event sourced entity subprotocol is still compatible across current and previous protocol versions.

pvlugter:
Since this is mainly just reorganisation and additional tests, I'll merge now. Reviews, questions, comments all still welcome of course.

PR #503, opened 2020-12-22
Add initial node support TCK implementation

pvlugter:
Begin to implement the new TCK for node support, starting with the event sourced TCK model. The node support event sourced implementation doesn't emit events immediately, so this currently applies the event handler to the command-local state as well, so the updated state can be returned in the response. And because events are not emitted immediately, there's no need to add the restart option for emit-then-fail, so the TCK is updated to test failures with or without the restart. See also #375.

I'll follow up later with the other TCK models, publishing a docker image, enabling for native image TCK, and so on.


pvlugter:
Not having any luck with the Travis build, keeps getting stuck without any progress at different points.

PR #515, opened 2021-01-22
Get GraalVM container images from GitHub packages

raboof:
Since they have disappeared from Docker Hub:
https://www.graalvm.org/docs/getting-started/#graalvm-community-container-images

May need an update to CircleCI to set up credentials to log in to
GitHub, let's see.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -84,7 +84,7 @@ jobs:
     - run:
         name: Build sbt native image
         command: |
-          docker login https://docker.pkg.github.com -u cloudstatebot -p ${GITHUB_TOKEN}
+          docker login https://docker.pkg.github.com -u cloudstatebot -p $GITHUB_TOKEN
           sbt "dockerBuildNativeCore publishLocal"
     - save_docker_image:
         tar_file: cloudstate-proxy-native-core.tar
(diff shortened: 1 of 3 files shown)


pvlugter:
I've updated the GITHUB_TOKEN env var to the token currently used for cloudstatebot, and reverted back the latest commits. But realising that the secrets will of course not be available to pull requests from forks. If we need to authenticate with secrets, then we'll likely need a different approach here, as we shouldn't enable secrets for forked pull requests.

pvlugter:
Ok, looking at this more closely. So the older version of Graal that we use is published to `docker.pkg.github.com/graalvm/container/community` but `docker.pkg.github.com` requires authentication for all pulls. The newer Graal docker images are at `ghcr.io/graalvm/graalvm-ce`, which allows unauthenticated pulls, but doesn't have the older version that we're using. Trying an upgrade to `20.3.1` and it doesn't build — will need changes. So possibilities as I see it:

- create some public guest credentials that can go into CI build directly
- republish the docker image for the older version to another repository
- update to a newer Graal version so we can pull from ghrc.io instead (this is what we'll want to do at some point anyway)

pvlugter:
I'm trying a minimal Graal update in #519

pvlugter:
Closing this one, with #519 passing tests.

raboof:
Great, thx!

PR #516, opened 2021-01-25
Update Akka HTTP 10.2.3

raboof:
I think we should not merge this before we have tested that the native image also still works for this version.

Building the native image on CI is currently broken, see #515


pvlugter:
@raboof, should be able to rebase this now, with CI tests working a bit better again.

Issue #411, opened 2020-08-19
Investigate benefits of upgrading & possibly upgrade to GraalVM 20.2


viktorklang:
https://www.graalvm.org/release-notes/20_2/#2020

viktorklang:
This version seems to address this issue that we were seeing: https://github.com/oracle/graal/issues/1306#issuecomment-516064882

wololock:
@viktorklang, try out `GraalVM CE 20.2.0 (build 11.0.8+10-jvmci-20.2-b03)` specifically. I did run my Groovy examples using the newest GraalVM CE 20.2.0 (build 25.262-b10-jvmci-20.2-b03, mixed mode) - the OpenJDK 8u262 one, and it still does not work. I hope it works for you :)

viktorklang:
Thanks for the tip, @wololock! :+1:

viktorklang:
Worth tracking: https://github.com/oracle/graal/issues/2761

viktorklang:
This issue of ours seem to have been recently closed:

https://github.com/oracle/graal/issues/2736
https://github.com/oracle/graal/issues/2369

viktorklang:
Graalvm 20.3 has been released: https://www.graalvm.org/release-notes/20_3/

pvlugter:
Minimal upgrade to 20.3.1 in #519. Leaving this issue open for any follow-ups and looking at upgrading to version 21.0.

PR #519, opened 2021-01-26
Update to Graal 20.3.1 and use github container registry

pvlugter:
Alternative to #515. Use `ghcr.io/graalvm/graalvm-ce` for docker images, and update to `20.3.1` (closest version published there). A couple of options had errors, which I've just commented out for now. Let's see how the integration tests go.


pvlugter:
Ok, this is building successfully and passing the integration tests now. I'll merge now to unblock CI, and we can continue looking into the Graal upgrade via #411.

PR #524, opened 2021-01-28
Update native image TCK integration tests

pvlugter:
Add a setting for whether to docker pull in TCK integration tests, so we have this on for TCK implementation docker images, and off for native image proxy so it uses the locally built one. Currently pulling a published image instead in #511 which is failing.

Also try enabling the node-support TCK in the native image integration tests.


pvlugter:
I'll merge this and keep an eye on CI builds. Since the native image cassandra smoke test fails regularly, we may need to disable that until it's looked at more closely.

PR #525, opened 2021-02-01
Disable git clone depth in travis

pvlugter:
Travis has a default git clone depth of 50, so when the most recent tag is further than 50 commits away then the version for latest docker images will be `0.0.0`. This should disable the clone depth so we always have a version in the docker images.


pvlugter:
The native image for cassandra smoke test fails so often, I think I'll disable it and create an issue.

PR #529, opened 2021-02-02
More specific NPE in registerEventSourcedEntity

ennru:
If the `findServiceByName` fails the user ends up with NPE.
This makes the exception point to the most likely wrong name in the lookup.


pvlugter (approved):
LGTM

PR #531, opened 2021-02-04
Fix unsafe state access in value entity proxy

raboof:
The 'handler' parameter to 'performAction' was
called from a future outside of message processing.

While it 'maybe never lead to problems in practice',
accessing actor state like that is generally not allowed,
and could have lead to confusing situations (e.g.
misleading 'sender' parameters or actual problems when
we'd refactor things later).

For this reason I didn't find a way to introduce a test
that covers the problem.


ralphlaude:
LGTM!

pvlugter (approved):
Good catch, LGTM.

Issue #536, opened 2021-02-18
Additional conditions not managed properly when reconciling StatefulService


marcoderama:
@jroper observes:

I created a StatefulService manually with kubectl, the resource was created successfully, however, when the operator tried to reconcile it, this exception was thrown:

```
2021-02-01T23:35:43.929Z	ERROR	controller	Reconciler error	{"reconcilerGroup": "cloudstate.io", "reconcilerKind": "StatefulService", "controller": "statefulservice", "name": "crdts", "namespace": "3e8cd683-4c57-418e-9d56-cbfc665dd90c", "error": "StatefulService.cloudstate.io \"crdts\" is invalid: status.conditions.lastTransitionTime: Invalid value: \"null\": status.conditions.lastTransitionTime in body must be of type string: \"null\""}
github.com/go-logr/zapr.(*zapLogger).Error
	/workspace/vendor/github.com/go-logr/zapr/zapr.go:128
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:237
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:209
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker
	/workspace/vendor/sigs.k8s.io/controller-runtime/pkg/internal/controller/controller.go:188
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156
k8s.io/apimachinery/pkg/util/wait.JitterUntil
	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:133
k8s.io/apimachinery/pkg/util/wait.Until
	/workspace/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:90
```

This was when creating a StatefulStore with no store type specified.

PR #537, opened 2021-02-18
Deal with additional conditions properly

marcoderama:
These changes are a fix for Issue https://github.com/cloudstateio/cloudstate/issues/536.

The fundamental diffs are
* ~~Make the Status component optional in both StatefulStores and StatefulServices.  This is the natural condition when they're first created.~~
* Fix up `ReconcileStatefulServiceConditions()`

After these changes, and testing with the setup of Issue https://github.com/cloudstateio/cloudstate/issues/536, the StatefulService entity ends up with a Status section looking like:
```
status:
  conditions:
  - lastTransitionTime: "2021-02-18T01:15:21Z"
    message: Stateful store no-store not supported
    reason: StoreNotSupported
    status: "True"
    type: NotReady
  - lastTransitionTime: "2021-02-18T01:25:22Z"
    message: Application is in the process of updating
    reason: UpdateInProgress
    status: "True"
    type: NotReady
```


marcoderama:
Tests now pass but converting to draft because I want to reconsider making the Status component optional.

